PR #3, opened 2018-10-23
Import subgraph from https://gitlab.com/bohendo/alchemy

shivgupt:
Sorry, I didn't have time to include the build process since it came from a monorepo and I didn't want to re-do a bunch of stuff that's already done.

Looking at Bo's repo might be helpful & avoid duplication of devops effort.


schema.graphql
```diff
@@ -1,8 +1,95 @@
...
+    stakes: [Stake!] @derivedFrom(field: "accountId")
+    votes: [Vote!] @derivedFrom(field: "accountId")
+    proposals: [Proposal!] @derivedFrom(field: "proposer")
+    redemptions: [Redemption] @derivedFrom(field: "accountId")
+}
+
+type DAO @entity {
+    avatarAddress: ID!
+    controllerAddress: Bytes
+    members: [Account!] @derivedFrom(field: "daoAvatarAddress")
```
orenyodfat: daoAvatarAddress should be avatarAddress ?

schema.graphql
```diff
@@ -1,8 +1,95 @@
...
+    reputation: BigInt
+    tokens: BigInt
+    stakes: [Stake!] @derivedFrom(field: "accountId")
+    votes: [Vote!] @derivedFrom(field: "accountId")
+    proposals: [Proposal!] @derivedFrom(field: "proposer")
+    redemptions: [Redemption] @derivedFrom(field: "accountId")
+}
+
+type DAO @entity {
+    avatarAddress: ID!
```
orenyodfat: does address should not be bytes ?

schema.graphql
```diff
@@ -1,8 +1,95 @@
...
-  address: Bytes!
-  nChoices: Int!
-  organization: Bytes!
-  paramsHash: Bytes!
-  proposer: Bytes!
+    submittedTime: BigInt!
+    proposalId: ID!
+    proposer: Bytes!
+    beneficiaryAddress: Bytes!
+    daoAvatarAddress: Bytes!
```
orenyodfat: rename daoAvatarAddress to avatarAddres

orenyodfat: should not that be daoAvatarAddress: DAO! ?

src/mappings/ContributionReward.ts
```diff
@@ -0,0 +1,94 @@
...
+export { allocate_memory }
+
+import { Entity, Address, U256, Bytes, Value, store, crypto, ByteArray } from '@graphprotocol/graph-ts'
+
+import { NewContributionProposal, RedeemReputation, RedeemEther, RedeemNativeToken, RedeemExternalToken } from '../types/ContributionReward/ContributionReward'
+
+import { concat, updateRedemption } from '../utils'
+
+export function handleNewContributionProposal(event: NewContributionProposal): void {
+    let accountId = crypto.keccak256(concat(event.params._beneficiary, event.params._avatar)).toHex()
```
orenyodfat: why ? is that redundant here ?

src/mappings/ContributionReward.ts
```diff
@@ -0,0 +1,94 @@
...
+    proposal.setString('contributionDescriptionHash', event.params._contributionDescription.toHex())
+    proposal.setAddress('externalToken', event.params._externalToken)
+    //proposal.setU256('periodLength', event.params._rewards[4])
+
+    store.set('Proposal', event.params._proposalId.toHex(), proposal as Entity)
+}
+
+export function handleRedeemReputation(event: RedeemReputation): void {
+    let rewardType = new Uint8Array(1)
+    rewardType[0] = 2
```
orenyodfat: why assign 2 ?

src/mappings/ContributionReward.ts
```diff
@@ -0,0 +1,94 @@
...
+    proposal.setAddress('beneficiaryAddress', event.params._beneficiary)
+    proposal.setI256('reputationChange', event.params._reputationChange)
+    proposal.setString('contributionDescriptionHash', event.params._contributionDescription.toHex())
+    proposal.setAddress('externalToken', event.params._externalToken)
+    //proposal.setU256('periodLength', event.params._rewards[4])
+
+    store.set('Proposal', event.params._proposalId.toHex(), proposal as Entity)
+}
+
+export function handleRedeemReputation(event: RedeemReputation): void {
```
orenyodfat: please document this function ,
It is not clear what it is doing

src/mappings/DaoCreator.ts
```diff
@@ -0,0 +1,12 @@
+import 'allocator/arena'
+export { allocate_memory }
+
+import { Entity, Value, store, crypto, ByteArray } from '@graphprotocol/graph-ts'
+
+import { NewOrg } from '../types/DaoCreator/DaoCreator'
+
+export function handleNewOrg(event: NewOrg): void {
```
orenyodfat: how about adding a function to get dao members here ? 
by looking on the controller mint reputations events

src/mappings/Reputation.ts
```diff
@@ -0,0 +1,31 @@
+import 'allocator/arena'
+export { allocate_memory }
+
+import { Entity, Address, Value, store, crypto, ByteArray } from '@graphprotocol/graph-ts'
+
+import { Reputation, Mint, Burn } from '../types/Reputation/Reputation'
+import { concat } from '../utils'
+
+let avatar = Address.fromString('a3f5411cfc9eee0dd108bf0d07433b6dd99037f1')
```
orenyodfat: why it is hard coded ?

src/mappings/Reputation.ts
```diff
@@ -0,0 +1,31 @@
...
+    if (account == null) {
+        account = new Entity()
+        account.setAddress('address', event.params._to)
+        account.setString('accountId', accountId)
+        account.setAddress('daoAvatarAddress', avatar)
+        account.setU256('reputation', event.params._amount)
+    }
+    store.set('Account', accountId, account as Entity)
+
+    let dao = new Entity()
```
orenyodfat: why creating new dao here ?
I guess in this event of mint reputation we should add the member to already existing dao if this member does not exist and it has a positive rep.

src/mappings/UController.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import { Entity, Value, store, crypto, ByteArray } from '@graphprotocol/graph-ts'
+
+import { MintReputation, BurnReputation, MintTokens } from '../types/UController/UController'
+
+export function handleMintReputation(event: MintReputation): void {
+    // Tracking reputation minted per account in native rep tracker
+
+    let dao = new Entity()
+    dao.setAddress('controllerAddress', event.address)
```
orenyodfat: why new dao here ?

orenyodfat: we should add member to already existing dao here

src/mappings/genesisProtocol.ts
```diff
@@ -0,0 +1,122 @@
...
+export { allocate_memory }
+
+import { Entity, Address, U256, Bytes, Value, store, crypto, ByteArray } from '@graphprotocol/graph-ts'
+
+import { GPExecuteProposal, Stake, Redeem, RedeemDaoBounty, RedeemReputation,NewProposal, ExecuteProposal, VoteProposal } from '../types/GenesisProtocol/GenesisProtocol'
+
+import { concat, updateRedemption } from '../utils'
+
+export function handleNewProposal(event: NewProposal): void {
+    let accountId = crypto.keccak256(concat(event.params._proposer, event.params._avatar)).toHex()
```
orenyodfat: does the proposalId is not a decent unique id ?

src/mappings/genesisProtocol.ts
```diff
@@ -0,0 +1,122 @@
...
+    stake.setU256('prediction', event.params._vote)
+    stake.setU256('stakeAmount', event.params._amount)
+
+    store.set('Stake', uniqueId, stake as Entity)
+}
+
+export function handleGPExecuteProposal (event: GPExecuteProposal): void {
+    let proposal = new Entity()
+
+    //proposal.setInt('state', event.params._executionState as u32)
```
orenyodfat: why it is comment out ?

schema.graphql
leviadam: avatarAddress

schema.graphql
shivgupt: Avatar address can be uses to uniquely identify the DAO hence I used ID.

src/mappings/genesisProtocol.ts
shivgupt: I have been debugging this one. State is uint8 type and the way it is handled in web assembly is giving garbage values this way.

src/mappings/UController.ts
shivgupt: Since dao entity has avatar as the unique identifier, it will update only the controllerAddress of an existing dao and not create a new dao if it already exists.
Computation to get entity from store and then updating controllerAddress if not already set is more than just setting the controllerAddress.

schema.graphql
orenyodfat: All addresses should be Address type (instead of Bytes)
https://github.com/graphprotocol/graph-ts/blob/25591ce9384ce9428e7b789b551e182ba4587596/index.ts#L144

src/mappings/UController.ts
shivgupt: > we should add member to already existing dao here

Reputation minted to founder member only emits mint event from the reputation contract. So adding members here was missing the founder members, hence I add the members as part of mint event in reputation contract

orenyodfat: > Since dao entity has avatar as the unique identifier, it will update only the controllerAddress of an existing dao and not create a new dao if it already exists.
> Computation to get entity from store and then updating controllerAddress if not already set is more than just setting the controllerAddress.

why to update controller address on the mint reputation event ?

orenyodfat: > > we should add member to already existing dao here
> 
> Reputation minted to founder member only emits mint event from the reputation contract. So adding members here was missing the founder members, hence I add the members as part of mint event in reputation contract

Though the Mint /Burn events of the Reputation contracts does not give you the information of which DAO this rep is belonging. You should do that via the U/Controller events

src/mappings/genesisProtocol.ts
orenyodfat: Oh. I see . thanks 👍

src/mappings/genesisProtocol.ts
orenyodfat: does u8 does not work ?

orenyodfat:
Lets use Address type for addresses (https://github.com/graphprotocol/graph-ts/blob/25591ce9384ce9428e7b789b551e182ba4587596/index.ts#L144)

schema.graphql
shivgupt: > All addresses should be Address type (instead of Bytes)
> https://github.com/graphprotocol/graph-ts/blob/25591ce9384ce9428e7b789b551e182ba4587596/index.ts#L144

I agree and I'm updating it to Address instead of bytes. But keeping it as ID wherever it is unique identifier, since it is serialized in same way just makes it explicit that the field is a unique identifier

src/mappings/ContributionReward.ts
shivgupt: This was missing the account entry. I just updated it. I am creating account for all users and then member accounts will be the ones with reputation

orenyodfat: should you just add that to the proposals entry ? The proposal and crProposals .

schema.graphql
shivgupt: Oops. Address in not a valid type for graphql schema. Its an ethereum type.

src/mappings/Reputation.ts
shivgupt: Ideally this address should dynamically be replaced by the the address of the avatar whose native reputation was specified in yaml. But since we did not have any auto build currently I had a hard coded address of the avatar corresponding to the reputation contract specified in the yaml.
I'm still unsure how this would fit for a base cache layer

tsuberim:
Moved over to https://github.com/daostack/subgraph/pull/5

PR #4, opened 2018-10-24
Devops

tsuberim:
- Generate abis from `@daostack/arc` automatically as part of `codegen`
- Allow configuration of the project to `mainnet`/`development` using yaml templating.
- Set up integration testing (unfortunately doesn't work right now due to bug in `ganache`)
- Development with or without `docker` & `docker-compose`.


orenyodfat:
@tsuberim which bug in ganache you are referring to ?

tsuberim:
@orenyodfat The bug is https://github.com/trufflesuite/ganache/issues/907 which has been fixed but not yet released as a new version.

tsuberim:
@ben-kaufman just to be clear I only mean the ABI interface definitions (not the whole `.json` compiled file). Contract addresses are configured in `ops/config.yaml`. Since ABIs are the same regardless of the chain, I don't see what you mean.

orenyodfat:
> @orenyodfat The bug is [trufflesuite/ganache#907](https://github.com/trufflesuite/ganache/issues/907) which has been fixed but not yet released as a new version.
Is that effect graph-node code or the code added for this pr ? Is there a way to bypass that ?
@tsuberim

orenyodfat:
> @ben-kaufman just to be clear I only mean the ABI interface definitions (not the whole `.json` compiled file). Contract addresses are configured in `ops/config.yaml`. Since ABIs are the same regardless of the chain, I don't see what you mean.

Please note that the deployment addresses can potentially
be fetched out from the contract.json . (e.g arc.js package currently contain that though I guess it will be changed in the future ..need to discuss) 
later on we can use this as part of the automate codeine (generate the yaml files)

orenyodfat:
@tsuberim so it is basically a graph-node issue ? https://github.com/graphprotocol/graph-node/issues/375 ?

tsuberim:
@orenyodfat No, its ganache not following the spec (not setting the `to` field to `null` for contract creation tx)

ben-kaufman:
> @ben-kaufman just to be clear I only mean the ABI interface definitions (not the whole `.json` compiled file). Contract addresses are configured in `ops/config.yaml`. Since ABIs are the same regardless of the chain, I don't see what you mean.

I see that you import the ABI json files directly from Arc, while the version of the contract being used currently on the mainnet is different from the one in Arc, meaning the ABI is different than the ABI of the live contract.

orenyodfat:
> > @ben-kaufman just to be clear I only mean the ABI interface definitions (not the whole `.json` compiled file). Contract addresses are configured in `ops/config.yaml`. Since ABIs are the same regardless of the chain, I don't see what you mean.
> 
> I see that you import the ABI json files directly from Arc, while the version of the contract being used currently on the mainnet is different from the one in Arc, meaning the ABI is different than the ABI of the live contract.

Right . we need to put some thoughts on how to handle multiple versions of the same contracts (with different abi or not)

README.md
```diff
@@ -1,22 +1,51 @@
...
+# DAOstack subgraph
+
+The DAOstack caching layer built on top of [TheGraph](https://thegraph.com/)
 
 # Get started
 
-Develop locally:
+## Configure the project
+
+### .env variables
```
orenyodfat: please add an .env template file

orenyodfat:
I guess it will be nice to add ganache-cli to package.json and have a script to run it with the mnemonic

tsuberim:
[Setup API/Integration testing framework](https://trello.com/c/BTS2vp1N/9-setup-api-integration-testing-framework)

orenyodfat:
Should we add ganache-cli to package.json so it can be run manually regardless docker ?
please update readme on how to run development + tests

README.md
```diff
@@ -1,22 +1,60 @@
-# daostack-subgraph
-A DAOstack subgraph for graph-node
+# DAOstack subgraph
 
-# Get started
+DAOstack subgraph for [TheGraph](https://thegraph.com/) project.
```
orenyodfat: Add a link to graph-node GitHub as theGraph

orenyodfat: Maybe "DAOstack subgraph using graph-node project."

PR #6, opened 2018-10-30
Reputation

tsuberim:
Support for the reputation contract.


leviadam:
Hi @tsuberim 
Is this ready for review? Or are you still working on it?

tsuberim:
@leviadam I think its ready for review.

ops/index.js
```diff
@@ -64,21 +63,14 @@ async function migrate(web3) {
...
-    data: DAOToken.bytecode,
-    arguments: ["GEN", "GEN", "1000000000000000000000000000"]
-  }).send();
-
-  const GP = new web3.eth.Contract(GenesisProtocol.abi, undefined, opts);
-  const gp = await GP.deploy({
-    data: GenesisProtocol.bytecode,
-    arguments: [gen.options.address]
+  const Rep = new web3.eth.Contract(Reputation.abi, undefined, opts);
+  const rep = await Rep.deploy({
```
orenyodfat: why deploy ?

package.json
```diff
@@ -7,6 +7,9 @@
     "type": "git",
     "url": "https://github.com/daostack/subgraph.git"
   },
+  "engines": {
+    "node": "8"
```
orenyodfat: should be 10.30.0 (recommended)

package.json
```diff
@@ -47,7 +50,7 @@
     "shell-exec": "^1.0.2",
     "ts-jest": "^23.10.4",
     "typescript": "^3.1.3",
-    "web3": "^1.0.0-beta.36",
+    "web3": "1.0.0-beta.36",
```
orenyodfat: is that not include in graph-cli ?

schema.graphql
```diff
@@ -1,8 +1,19 @@
-type Proposal @entity {
-  id: ID!
-  address: Bytes!
-  numOfChoices: BigInt!
-  organization: Bytes!
-  paramsHash: Bytes!
-  proposer: Bytes!
+type ReputationHolder @entity {
```
orenyodfat: should we have schema.graphql under directories (e.g /reputation ?)

test/integration/Reputation.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+        contract,
+        address,
+        amount
+      }
+    }`);
+
+    expect(reputationBurns.length).toEqual(3);
+    expect(reputationBurns).toContainEqual({
+      txHash: txs[2],
+      contract: reputation.options.address.toLowerCase(),
```
orenyodfat: can we set the address type to be lower case in the db by default ?

package.json
leviadam: Also better use `>=`

src/mappings/Reputation.ts
```diff
@@ -0,0 +1,54 @@
...
+import { concat, zero256, isZero } from '../utils'
+
+function update(contract: Address, owner: Address): void {
+    let rep = Reputation.bind(contract);
+    let ent = new Entity();
+    let id = crypto.keccak256(concat(contract, owner)).toHex();
+    ent.setString('id', id);
+    ent.setAddress('contract', contract);
+    ent.setAddress('address', owner);
+    let balance = rep.balanceOf(owner);
```
orenyodfat: Here it validate the state (balance) via the blockchain (which can be calculate off chain by adding /reducing balances.
Pros : 
1. Additional validation against the blockchain.
Cons: 
1. Redundant web3 call . (time) 

questions : 
- Should we consider adding the balance to the Mint/Burn events ?
- How the performance will be effected by reading the state from the bc ? let say you have multiple Mint/Burn events on the same block.

ops/index.js
tsuberim: This is the migration script

package.json
tsuberim: Don't know. But anyway we also use web3 for testing and migration. Since its still unstable we need to fix the version.

test/integration/Reputation.spec.ts
tsuberim: Its already lowercase in the db. Its mixed case in web3.

src/mappings/Reputation.ts
tsuberim: The main concern in my opinion is that we don't want to replicate the entire logic of the contract inside the cache. This would make the cache pretty brittle.
answers:
- I don't think we need to add the balance to the events.
- I don't know how `graph-node` actually implements those web3 calls, they may do it in parallel across mappings which means its not probably going to be a problem.

schema.graphql
tsuberim: I thought about this, yeah I think we probably want to modularize the graphql schema. This can easily be done via [handlebars partials](https://handlebarsjs.com/partials.html)

package.json
tsuberim: https://github.com/daostack/subgraph/pull/6/commits/be75e747f40475b2e3d8cdbba52fa4526f504742

schema.graphql
tsuberim: Implemented in: https://github.com/daostack/subgraph/pull/6/commits/f79f0d2650de81c95150b4000992ca90ec26c5e0 What do you think?

ops/index.js
orenyodfat: are we migrating here ? not sure .
If this is for testing purposes , so move it under test folder

package.json
orenyodfat: should be 10.13.0

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "8"
+  - "10"
```
leviadam:
```suggestion
  - ">=10"
```

src/mappings/Reputation.graphql
```diff
@@ -0,0 +1,19 @@
+type ReputationHolder @entity {
+    contract: ID!
+    address: ID!
+    balance: BigInt!
+}
+
+type ReputationMint @entity {
```
leviadam: 1. Are we sure we need to keep the history of mint/burn and not just the balance? Do you see a use case for that?
2. What about `totalSupply`?

tsuberim: `Reputation` doesn't have a `name` and `symbol`. we do need to add `totalSupply` though.

test/integration/Reputation.spec.ts
tsuberim: addresses are saved in lowercase in the db, eg. `0xa1b...`. web3 returns addresses in mixed case, eg. `0xa1B...`. we need to convert the web3 address to lowercase in order to match it to the one the api returns (which is lowercase).

schema.graphql
tsuberim: hmm.. yeah can be removed.

package.json
tsuberim: why does the minor version matter? just `10` is enough imo.

ops/index.js
tsuberim: this is the `migrate` function in `index.js` that is run before the tests. Not sure we need to move it to the `test` folder.

tsuberim:
@orenyodfat 
This is not the package's main index.js its in the `ops` directory (which lets you include it as `require('./ops')`).

PR #7, opened 2018-10-31
Ucontroller

tsuberim:
Support for UController.


ops/config.yaml
```diff
@@ -3,7 +3,7 @@ common: &common
 mainnet:
   <<: *common
   addresses:
-    GenesisProtocol: "0x8940442e7f54e875c8c1c80213a4aee7eee4781c"
+    Reputation: "0x13ad61aa8f695ce64711a51a6feccb48a275aaf8"
```
orenyodfat: Is this related to this PR ? UController ?

src/mappings/UController.graphql
```diff
@@ -0,0 +1,63 @@
+type UControllerOrganization @entity {
+    avatarAddress: ID!
```
orenyodfat: how about 'avatarAddress : Avatar! '?

src/mappings/UController.graphql
```diff
@@ -0,0 +1,63 @@
...
+    nativeToken: Bytes!
```
orenyodfat: nativeToken: NativeToken!

src/mappings/UController.graphql
tsuberim: Why do we need another entity? also Avatar is another contract.

src/mappings/UController.graphql
tsuberim: Why do we need another entity? also ERC20 is another contract.

ops/config.yaml
tsuberim: not related but needed.

orenyodfat: how it is needed ? 
lets do other config.yaml for the reputation..

src/mappings/UController.graphql
orenyodfat: this is the power of Graph(ql).
controller point to Avatar

tsuberim: But we have no Avatar entity right now

orenyodfat: ok. lets do it only for rep

PR #11, opened 2018-11-06
fix isZero

orenyodfat:
There was an index out of bounds in isZero because graph-node changed BigInt to be an array of bytes.

Now the rep test pass .
Need to add more 500 ms delay on tests.

### DIFF
--- a/test/integration/util.ts
+++ b/test/integration/util.ts
@@ -15,7 +15,7 @@ import * as HDWallet from "hdwallet-accounts";
 
 const { node_http, ethereum, test_mnemonic } = process.env;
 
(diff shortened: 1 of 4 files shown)


test/integration/util.ts
```diff
@@ -15,7 +15,7 @@ import * as HDWallet from "hdwallet-accounts";
 
 const { node_http, ethereum, test_mnemonic } = process.env;
 
-export async function query(q: string, maxDelay = 500) {
+export async function query(q: string, maxDelay = 1000) {
```
tsuberim: should be 500?

orenyodfat: for some reason it fail with 500

PR #16, opened 2018-11-07
ContributionReward

tsuberim:


test/integration/ContributionReward.spec.ts
```diff
@@ -0,0 +1,243 @@
...
+        const descHash = '0x0000000000000000000000000000000000000000000000000000000000000123';
+        const rewards = {
+            rep: 1,
+            nativeToken: 2,
+            externalToken: 3,
+            eth: 4,
+            periods: 5,
+            periodLength: 6
+        }
+        const propose = contributionReward.methods.proposeContributionReward(
```
orenyodfat: need await ?

tsuberim: no, that's only building the object which you can `call(..)` or `send(..)` on.
Below we do both and `await` the results.

tsuberim:
No longer WIP.

PR #18, opened 2018-11-11
Add subscription test

orenyodfat:
For now , a very simple event (Reputation Mint) subscription via web socket as a POC.
Using apollo client lib https://github.com/apollographql


test/integration/Subscription.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+          reputationMints {
+            contract
+            amount
+            address
+          }
+        }
+        `;
+
+      const subscriptionClient =await  createSubscriptionObservable(
+          'http://graph-node:8001/by-name/daostack', // GraphQL endpoint
```
tsuberim: should be `process.env.node_ws`

test/integration/Subscription.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+          SUBSCRIBE_QUERY,                                       // Subscription query
+          {address: accounts[0].address.toLowerCase()}                                                // Query variables
+        );
+
+      let event;
+
+      var consumer =await subscriptionClient.subscribe(eventData => {
```
tsuberim: I think its easier to do `await subscriptionClient.toPromise()` which just returns the first event as a promise.

test/integration/Subscription.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+            amount
+            address
+          }
+        }
+        `;
+
+      const subscriptionClient =await  createSubscriptionObservable(
+          'http://graph-node:8001/by-name/daostack', // GraphQL endpoint
+          SUBSCRIBE_QUERY,                                       // Subscription query
+          {address: accounts[0].address.toLowerCase()}                                                // Query variables
```
tsuberim: Where is this used inside the query?

test/integration/Subscription.spec.ts
orenyodfat: Not sure how

tsuberim: you just call `toPromise` on the observable.

test/integration/util.ts
```diff
@@ -60,3 +66,16 @@ export function padZeros(str: string, max = 36) {
     str = str.toString();
     return str.length < max ? padZeros("0" + str, max) : str;
 }
+
+
+const getWsClient = function(wsurl) {
```
tsuberim: maybe make these a single function `watchQuery(query: string): Observable` ?

tslint.json
```diff
@@ -0,0 +1,9 @@
+{
+    "defaultSeverity": "error",
+    "extends": [
+        "tslint:recommended"
+    ],
+    "jsRules": {},
+    "rules": {},
```
tsuberim: I like single quotes

orenyodfat:
"Should we add such tests for all contracts? seems like duplicated work."

@tsuberim the test is just to make sure the subscription is working.not sure we should do that for all mappings.

tslint.json
orenyodfat: This file is autogenerated by tslint init.
it also lint the code to use double quotes.
I do not care ..

tsuberim:
```
"rules": {
  "quotemark": [true, "single"]
}
```

tsuberim:
@orenyodfat So is this sufficient? Also does the event fire even when a property is changed in an already existent entity?

tslint.json
orenyodfat: lets argue about that :)

tsuberim: single quotes don't required pressing `shift`. Boom!

tslint.json
```diff
@@ -0,0 +1,14 @@
...
+    "defaultSeverity": "error",
+    "extends": [
+        "tslint:recommended"
+    ],
+    "jsRules": {},
+    "rules": {
+              "object-literal-sort-keys": false,
+              "only-arrow-functions": false,
+              "no-var-requires": false,
+              "prefer-const": false
```
tsuberim: single quote

PR #21, opened 2018-11-13
add proposal entity

orenyodfat:
high level proposal entity which hold both CR and GP proposal attributes.


tsuberim:
Please follow the "API only entites" section in the [DAOstack node spec](https://docs.google.com/document/d/1dqNks1qRXOgjlSfvoM80gEj2XwP1bV9rXlicVk6jjR4/edit?ts=5b9e1794#heading=h.9t2t3phc1mh1).

PR #22, opened 2018-11-13
WIP: A higher-level Schema

bohendo:
This PR mostly implements the [DAOstack node entities spec](https://docs.google.com/document/d/1dqNks1qRXOgjlSfvoM80gEj2XwP1bV9rXlicVk6jjR4/edit?ts=5bb27b80#heading=h.9t2t3phc1mh1).

Plus a few relatively small changes. Plus at least a few bugs. This branch is currently broken, don't merge it yet but please browse & start providing feedback if you have time, thanks!


bohendo:
The schema is probably the most interesting part:

https://github.com/daostack/subgraph/pull/22/files#diff-270b2728f0e47a3fcd2f0766d605f06c

.gitignore
```diff
@@ -1,10 +1,3 @@
...
-*.log
-subgraph.yaml
-docker-compose.yml
-config.json
-abis
-yarn.lock
-src/types
-schema.graphql
\ No newline at end of file
+*.swp
```
orenyodfat: why remove these?

package.json
```diff
@@ -34,7 +34,7 @@
   },
   "dependencies": {
     "@daostack/arc": "0.0.0-alpha.56",
-    "@graphprotocol/graph-cli": "./graphprotocol-graph-cli-0.4.0.tgz",
+    "@graphprotocol/graph-cli": "^0.4.0",
```
orenyodfat: this patch is here because https://github.com/graphprotocol/graph-cli/pull/165 which is not yet released.

orenyodfat:
@bohendo please rebase that on master so it will be easy to monitor and review the diffs

.gitignore
bohendo: A bug was introduced on my end because subgraph.yaml & schema.graphql were ignored & I didn't notice them not being committed. These definitely need to be removed from the .gitignore.

I got a little overexcited & shouldn't have removed dist or *.log or some of the others though, you're right

bohendo:
@orenyodfat

1. Right when I posted this PR it was again the existing shiv branch. But then I rebased it onto master so it should be good now. Are there changes you see that shouldn't be included in this PR?

2. Which aspects of the existing directory structure are missing? The mappings & schema are in `src` & build tools are in `ops`. Are you talking about the output folder being called `build` instead of `dist`?

3. Honestly, my #1 priority right now is fixing Alchemy's load time. It seems like adding the low level mappings come at the cost of making the graph node sync more slowly & bloating the database. What benefits to they provide to offset this cost & how would they help Alchemy load faster?

.gitignore
orenyodfat: subgraph.yaml & schema.graphql should not be in .gitignore .
it is generated automatically by https://github.com/daostack/subgraph/blob/master/subgraph.handlebars.yaml and other scripts .
Please read Readme ...on how to use the env

orenyodfat:
duplicate at #29

PR #23, opened 2018-11-14
Avatar caching

ben-kaufman:


ben-kaufman:
Ok, should be working now.
@orenyodfat  @tsuberim Please note the change I did to make the subscription test work as I'm not sure if the change doesn't damage the test (I think it should be fine)...

test/integration/Subscription.spec.ts
```diff
@@ -36,6 +36,8 @@ describe('Subscriptions', () => {
         );
 
         let event;
+
```
orenyodfat: you first need to subscribe and then trigger the event.

orenyodfat:
Still need to update util BigInt adding and sub function.
We can live without it for now (using small i32 numbers)..
I opened an issue https://github.com/daostack/subgraph/issues/24

PR #30, opened 2018-11-30
do not docker-ignore package-lock

jellegerbrandy:


.dockerignore
```diff
@@ -1,5 +1,4 @@
 node_modules
-package-lock.json
 *.log
```
jellegerbrandy: including package-lock in the docker file will make our builds slightly more predicable

Dockerfile
```diff
@@ -12,7 +11,7 @@ RUN chmod +x /wait /entry
 
 WORKDIR /usr/app
 COPY . .
-RUN npm install
+RUN npm ci
```
orenyodfat: what ci stand for ?

jellegerbrandy: "continuous integration"

jellegerbrandy: https://docs.npmjs.com/cli/ci

PR #32, opened 2018-12-04
improve migration script

jellegerbrandy:
Fix: #34


Dockerfile
```diff
@@ -11,7 +11,7 @@ RUN chmod +x /wait /entry
 
 WORKDIR /usr/app
 COPY . .
-RUN npm ci
+RUN npm install
```
orenyodfat: why is that changed back ?

jellegerbrandy: It is changed back because I'd like to buidl the container in the `node_modules/@daostack/subgraph` repo, but the intallation of npm does not include the `package-lock.json`. And so `npm ci` fails. Unfortunately, because it would be better/more predictable.

tsuberim: @jellegerbrandy We should instead publish the image to dockerhub and use it in the `docker-compose.yml` to both avoid this problem and shorten build times.

docker-compose.yml
```diff
@@ -8,7 +8,7 @@ services:
       - ganache
     environment:
       ethereum: http://ganache:8545
-      test_mnemonic: behave pipe turkey animal voyage dial relief menu blush match jeans general
+      test_mnemonic: myth like bonus scare over problem client lizard pioneer submit female collect
```
orenyodfat: just for curiosity ...why this was changed?

ben-kaufman: I think it's because the new one is the default mnemonic of Ganache.

Dockerfile
jellegerbrandy: Yes I agree.

docker-compose.yml
jellegerbrandy: This is the mnemonic that gnereated the accoutns of ganache when it runs in `--deterministic` mode but without inventing our own test mnemonic. https://github.com/daostack/devops#ganache

jellegerbrandy: (see https://github.com/daostack/subgraph/pull/32/files#diff-4e5e90c6228fd48698d074241c2ba760L50)

ops/deployDaoStack.js
```diff
@@ -0,0 +1,46 @@
+const DAOstackMigration = require('@daostack/migration');
```
jellegerbrandy: this script replaces `ops/migration.js` and uses the scripts from the mgration repo instead

ops/generate-subgraph.js
```diff
@@ -16,11 +17,17 @@ async function generateSubgraph() {
 	const dataSources = files.map(file => {
 		const contract = path.basename(path.dirname(file));
 		const { abis, entities, eventHandlers } = yaml.safeLoad(fs.readFileSync(file, 'utf-8'));
+
+		const contractAddress = addresses.private.base[contract] || addresses.private.dao[contract];
+
+		if (!contractAddress) {
+			throw Error(`Address for contract ${contract} not found in ${migrationFile}`);
```
jellegerbrandy: let's try to do explicit error handling in our scripts so we catch errors early

Dockerfile
jellegerbrandy: Actually, you are just using this container to run the tests in, right? (Is that a good idea?) In that case, there would be no need to build that container at all...

src/mappings/Avatar/mapping.ts
```diff
@@ -1,5 +1,5 @@
 import 'allocator/arena';
-export { allocate_memory };
+//// export { allocate_memory };
```
orenyodfat: please remove this comments

src/mappings/Avatar/mapping.ts
```diff
@@ -18,7 +18,7 @@ function handleAvatarBalance(
 
   let avatar = store.get('AvatarContract', address.toHex()) as AvatarContract;
   if (avatar == null) {
-    avatar = new AvatarContract();
+    avatar = new AvatarContract(address.toHex());
     avatar.id = address.toHex();
```
orenyodfat: remove redundant.id - this could be done in other pr .

ops/deployDaoStack.js
```diff
@@ -0,0 +1,48 @@
...
+  options = {
+    // web3 provider url
+    provider,
+    quiet: true,
+    // disable confirmation messages
+    force: true,
+    // filepath to output the migration results
+    output: path.resolve(migrationFileLocation),
+    // private key of the account used in migration (overrides the 'mnemonic' option)
+    // privateKey: '0x8d4408014d165ec69d8cc9f091d8f4578ac5564f376f21887e98a6d33a6e3549',
```
orenyodfat: do we need this comment here ? @jellegerbrandy

jellegerbrandy: no we do not :-)

PR #37, opened 2018-12-18
Update subgraph manifest to new specification format

davekaj:
This pull request is just updating the manifest of the subgraph to add in new fields. We are using these new fields to point to the subgraph, so we forked it and made the updates.

We would love for it to be merged into the master branch of the DAOStack subgraph.

### DIFF
--- a/ops/generate-subgraph.js
+++ b/ops/generate-subgraph.js
@@ -47,7 +47,7 @@ async function generateSubgraph() {
 	const subgraph = {
 		specVersion: '0.0.1',
 		schema: { file: './schema.graphql' },
-		description: `DAOstack enables the effective governance of self-organizing collectives. Find DAO's members, proposals, votes, and more.`,
+		description: `DAOstack (https://daostack.io) powers decentralized companies, funds and markets to make fast and innovative decisions at scale.`,
 		repository: 'https://github.com/daostack/subgraph',
 		dataSources,
 	};
--- a/ops/generate-subgraph.js
+++ b/ops/generate-subgraph.js
@@ -47,7 +47,7 @@ async function generateSubgraph() {
 	const subgraph = {
 		specVersion: '0.0.1',
 		schema: { file: './schema.graphql' },
-		description: `DAOstack (https://daostack.io) powers decentralized companies, funds and markets to make fast and innovative decisions at scale.`,
+		description: `This subgraph provides an indexing and caching layer for DAOstack Arc, which is the base layer of the DAO stack.`,
 		repository: 'https://github.com/daostack/subgraph',
 		dataSources,
 	};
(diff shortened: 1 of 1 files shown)


davekaj:
@bohendo

bohendo:
I'd love to help get this merged but unfortunately I don't have permission to do so, @tsuberim or @orenyodfat could probably help out though

ops/generate-subgraph.js
```diff
@@ -46,6 +47,8 @@ async function generateSubgraph() {
 	const subgraph = {
 		specVersion: '0.0.1',
 		schema: { file: './schema.graphql' },
+		description: `DAOstack enables the effective governance of self-organizing collectives. Find DAO's members, proposals, votes, and more.`,
```
orenyodfat: how about replace this description with https://daostack.io ?

ops/generate-subgraph.js
```diff
@@ -27,6 +27,7 @@ async function generateSubgraph() {
 		return {
 			kind: 'ethereum/contract',
 			name: `${contract}`,
+			network: 'mainnet',
```
orenyodfat: I guess that 'mainnet' should be replaced with '${network}' as the subgraph suppose to run not just over mainnet .

davekaj: Yeah that makes a lot of sense. @bohendo I looked around briefly and I couldn't figure out the best place to get a `${network}` variable to use in the codegen. Do you have any suggestions? Maybe getting it when the `migration.json` file is created?

ops/generate-subgraph.js
davekaj: I updated the description to the same on one on the front page of the website, and added in the URL. The description field is representative of people from the outside looking in, who want to know what DAOstack is, so that they can decide if they are interested in the data.

ops/generate-subgraph.js
```diff
@@ -47,7 +47,7 @@ async function generateSubgraph() {
 	const subgraph = {
 		specVersion: '0.0.1',
 		schema: { file: './schema.graphql' },
-		description: `DAOstack enables the effective governance of self-organizing collectives. Find DAO's members, proposals, votes, and more.`,
+		description: `DAOstack (https://daostack.io) powers decentralized companies, funds and markets to make fast and innovative decisions at scale.`,
```
orenyodfat: - I guess this description field should describe the repo(subgraph) not the project who own/use it (DAOstack) .no? 
Any how i will replace companies with  organisations.

davekaj: Yes I can see what you are saying. I have updated it to reflect this better.

ops/generate-subgraph.js
ben-kaufman: @davekaj I would suggest adding that to `settings.js`.

PR #44, opened 2018-12-28
update migration

jellegerbrandy:


ben-kaufman:
I think we should wait to first have here the latest version of the migration which has the test cases.

ben-kaufman:
Should use this version of migration: `0.0.0-alpha.58-v9`.

PR #57, opened 2019-01-10
do not depend on setTimeout in subscription test

jellegerbrandy:


test/integration/Subscription.spec.ts
```diff
@@ -54,8 +56,8 @@ describe('Subscriptions', () => {
 
     await reputation.methods.mint(accounts[4].address, '99').send();
 
-    // //wait a second
-    await new Promise((res) => setTimeout(res, 1000));
+    // wait until the subscription callback has been called
+    await waitUntilTrue(() => nextWasCalled);
```
orenyodfat: the test has a timeout of 2500 ms.
it can exit without unsubscribe ..this may lead to failure on successive subscriptions.

orenyodfat: yea,  you are right about failure on successive subscription ... though the test can still fail on a timeout ..no ?

jellegerbrandy:
closing this - has become obsolete

PR #70, opened 2019-01-20
use migration repo docker

### LINKED ISSUE #68: use the docker-hub image from the migration repo for the ganache container
We've got this nice pre-built container, let's use it. 
https://hub.docker.com/r/daostack/migration/tags

ben-kaufman:
fix #68


jellegerbrandy:
(If we use the rc.6 docker container, we should also upgrade the @daostack/migration reference in package.json to rc6.)

Issue #67, opened 2019-01-15
proposalVote refers to the member entity, but the corresponding member may not exist (anymore)


jellegerbrandy:
The reputation holder may loose all of her rep after voting, and in that case, there would not be a member anymore.

Better to just have a `voter` field that is a simple address.

This issue is similar to #60

jellegerbrandy:
(also because it is easier to use: go from `address --> get votes of address` instead of `address --> get member.id of address --> get votes of member.id`)

Issue #60, opened 2019-01-11
rewards entity refers to the member entity, but the beneficiary may not be a member of the  DAO 


jellegerbrandy:
This (probably, I cannot check bc of issue #59) means that if the reward goes to a non-member, this field will be null (?)

It is probably better to not change the entity and have a simple "beneficiary" field that contains the address of the beneficiary.

PR #71, opened 2019-01-21
Change use of member to address

ben-kaufman:
fix: #60 
fix: #67 
fix: #72


src/domain/member.ts
```diff
@@ -64,6 +64,8 @@ export function updateMemberTokens(
   let dao = getDAO(daoAddress.toHex());
   let token = DAOToken.bind(hexToAddress(dao.nativeToken));
   let member = getMember(address, daoAddress);
-  member.tokens = token.balanceOf(address);
-  saveMember(member);
+  if (member.reputation.toString().charAt(0) !== '0') {
```
orenyodfat: is there a  case where updateMemberToken is called where the rep is 0?

ben-kaufman: Yes, I think this function is called on every token transfer so it might not actually be between members. That's more of a bugfix.

ben-kaufman:
Added tests

orenyodfat:
Not all tests are passing ...is that not deterministic ?

ben-kaufman:
@orenyodfat I'm pretty sure it is deterministic. Travis tests sometimes fail for no clear reason but pass after re-run. Not even sure how to debug this as it happens only in Travis and not very often.

orenyodfat:
> @orenyodfat I'm pretty sure it is deterministic. Travis tests sometimes fail for no clear reason but pass after re-run. Not even sure how to debug this as it happens only in Travis and not very often.

Ok. lets open an issue for that..so we can track it in the future.

PR #84, opened 2019-01-31
fix deployment error from issue 82

### LINKED ISSUE #82: deployment of graph-node sometimes fails with a CRIT error
Steps to reproduce (unreliable):

```
docker-compose up graph-node
npm run deploy // (in another terminal) this will work fine
docker-comoose down -v
docker-compose up graph-node
npm run deploy // (again in the second terminal)
```
The last step will, often but not always, fail with the following errors in the grapn-node console:

```
graph-node_1          |  Jan 31 07:45:37.436 INFO Received subgraph_create request, params: SubgraphCreateParams { name: SubgraphName("daostack") }
graph-node_1          | Jan 31 07:45:41.183 CRIT failed to resolve subgraph name to ID: ErrorMessage { msg: "Subgraph entity without `currentVersion`" }, location: src/libcore/result.rs:1009
alchemy_graph-node_1 exited with code 1
```

jellegerbrandy:
Should fix #82 (but does not, yet)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,8 @@
 version: '3'
 services:
   graph-node:
-    image: graphprotocol/graph-node:latest@sha256:436846910aacbdd81250add747a4be3346e24a460b774c3514025e29f05f0a2f
+    image: graphprotocol/graph-node:latest@sha256:bc64577f8ed7d2a7badb575d85d99bf017bdddd68bf20b032c184654d27a4bef
+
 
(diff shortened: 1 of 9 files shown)


ben-kaufman:
Did you manage to fix it? (honestly, this bug is freaking me out already)

jellegerbrandy:
The graphnode guys in the chat said that this should fix it... I hope they were right :-)

Issue #43, opened 2018-12-28
implement "tokensNeededToPreBoost"


jellegerbrandy:
as a property of the `proposal` entity. 
Alchemy shows "how much rep is needed to boost this proposal", and this  property should give us the threshold we need to reach to boost the proposal.

jellegerbrandy:
this does not really address the original issue. But there are some doudbts now about the funciontality that alchemy earth actually needs, I'll re-open this issue if we still need it.

tibetsprague:
ok so looks fine to calculate GENs left to boost , as long as we know the constant. does that constant come from the DAO? is it always the same?

jellegerbrandy:
The "alphaconstant" that Oren mentinos is actually called "thresholdConst"  https://github.com/daostack/infra/blob/master/contracts/votingMachines/GenesisProtocolLogic.sol#L34 (can you confirm, Oren?)

I'm not sure what the best practice is here - in my view it makes sense to index these parameters in the subgraph, but there is no real place. In any case, for now, we can get it from the contract itself directly.

tibetsprague:
i think constants like this should be indexed on the DAO, since they can be customized per dao

jellegerbrandy:
Ok, I'm reopening this AGAIN :-) because calculating it on the client and then keeping it in sync is a lot of work. So what we want is an attribute that perhaps is best called: `tokensNeededToPreBoost`, and given the formula: 
 `(S+/S-) > AlphaConstant^NumberOfBoostedProposal.`

we would get `tokensNeededToPreBoost = (alphaconst*NumberOfBoostedProposals * StakesAgainst) - StakesFor`

orenyodfat:
I guess that alchemy would like to know : 
- upStake tokens needed to preBoost a Qued proposal.
- downStake tokens needed to Qued proposal from preBoosted.
How about the client will read the proposal entity from subgraph which will include current S+ and S- and proposal.confidenceThreshold (for the case of preBoosted)...and will do the calculation it self? 
The dao threshold can be calculated on the client side as well -alpha.power(NumberOfBoostedProposal) . so it need to know the number of current boosted proposal..and i guess it can have that,
@jellegerbrandy

jellegerbrandy:
[...]

PR #86, opened 2019-02-03
Added proposal confidence

ben-kaufman:
Fix: #43


ben-kaufman:
@jellegerbrandy does this solves the ordering of proposals in the client or do you need anything else here?

Issue #55, opened 2019-01-10
Need a way to get the last token Approval event


jellegerbrandy:
Alchemy needs to now how many tokens are currently approved. There is no way I can see to get that info fromt he subgraph. If te `TokenApproval` object is unique for a user/token pair, we can get that, but if more than 1 approval event happened, we need to get the last one, and there is no way to see which one that is.

Probably the most elegant solution would be if we add an allowances array to the TokenHolder entity that maps spenders to amounts. I.e. something like:

```
  {
        "balance": "1000000000000000000000",
        "contract": "0x79183957be84c0f4da451e534d5ba5ba3fb9c696",
        "id": "0x6abc6f657694652f91d04c35238727d9b414b640284835f69bc9e780c0efe11f",
        "allowances": {
             "0xAddressOfSpender" : 1234
        }
      }
```

ben-kaufman:
This should track allowance to the address of the Genesis Protocol right?

jellegerbrandy:
Well, I would keep it general, and just track all allowances, i.e. just save all the  `Allowance` event calls (I don't remember what the event is). In noraml usage in alchemy, the spender is the  genesisprotocol, but I see no reason to not also track the others...

jellegerbrandy:
Reopening this. I think the "allowances" entity should also contain the token contract address that the allowance pertains to. This is important because of future-proofing (we will have to monitor different otkens in the future)

jellegerbrandy:
Another thing that is issing is the possibility to query allowances by owner address and spender (i.e. the typical case), 
in the current implementation, the "owner" field on the allowances entity refers to a TokenHolder. This is (1) wrong (an account could have 0 token balance but will have an allowance) and (2) very inconvenient for querying (I see no way to get the allowance of an account for a certain spender in a single graphql query)

Solution: just use owner address, not a TokenHolder entity.

jellegerbrandy:
closing this and making a new issue dedicated specific to fixing he allowance entity #101

PR #88, opened 2019-02-04
Added allowances

ben-kaufman:
Fix: #55

### DIFF
--- a/src/mappings/DAOToken/mapping.ts
+++ b/src/mappings/DAOToken/mapping.ts
@@ -1,6 +1,6 @@
 import 'allocator/arena';
 
-import { Address, BigInt, crypto, store } from '@graphprotocol/graph-ts';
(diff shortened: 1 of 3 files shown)


src/mappings/DAOToken/mapping.ts
```diff
@@ -51,6 +52,10 @@ export function handleTransfer(event: Transfer): void {
   ent.value = event.params.value;
 
   store.set('TokenTransfer', ent.id, ent);
+
+  if (event.params.from !== event.transaction.from) {
+    updateAllowance(event.address, event.params.from, event.transaction.from);
```
orenyodfat: why it is under the Transfer event

ben-kaufman: Because a transfer may reduce allowance if it was a transfer from

orenyodfat:
Is this suppose to handle GEN allowance for staking ? 
GEN is not DAOToken.

PR #104, opened 2019-02-17
Added tokensNeededToPreBoost

ben-kaufman:
Fix: #43

### DIFF
--- a/src/domain/schema.graphql
+++ b/src/domain/schema.graphql
@@ -72,7 +72,7 @@ type Proposal @entity {
 	queuedVotePeriodLimit: BigInt!
 	boostedVotePeriodLimit: BigInt!
 	preBoostedVotePeriodLimit: BigInt!
-	thresholdConst: BigInt!
+	thresholdConst: Float!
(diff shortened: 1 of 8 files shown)


src/domain/schema.graphql
```diff
@@ -64,6 +65,7 @@ type Proposal @entity {
 	stakesFor: BigInt!
 	stakesAgainst: BigInt!
 	confidence: BigInt!
+	tokensNeededToPreBoost: BigInt
```
orenyodfat: add downStakeTokenToShiftBackToQ

orenyodfat: please handle ConfidenceLevelChange event https://github.com/daostack/infra/blob/master/contracts/votingMachines/GenesisProtocolLogic.sol#L554

orenyodfat:
Due to assembly script limitation supporting floating points numbers...we decided to do these calculation on the client library side.
Subgraph will provide the upStake ,downStake and confidenceThreshold of a proposal .
The dao threshold will be added to the DAO entity ,which is already linked to the proposal entity.

PR #105, opened 2019-02-17
Fix allowances

### LINKED ISSUE #101: Allowance entity should index owner and spender *address*
The allowance entity current refers to the owner and spender as a tokenholder enitty. It is easier if we just refer to owner and spender by address.

See the end of the discussion on this issue: https://github.com/daostack/subgraph/issues/55

### LINKED ISSUE #106: allowance entity also needs a `contract` property
so we know which token the allowance is referring to

ben-kaufman:
Fix #101 
Fix #106

### DIFF
--- a/src/mappings/DAOToken/schema.graphql
+++ b/src/mappings/DAOToken/schema.graphql
@@ -11,12 +11,12 @@ type TokenHolder @entity {
 	contract: Bytes!
 	address: Bytes!
 	balance: BigInt!
-	allowances: [Allowance!] @derivedFrom(field: "owner")
 }
 
 type Allowance @entity {
 	id: ID!
-	owner: TokenHolder!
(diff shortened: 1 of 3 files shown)


src/mappings/DAOToken/schema.graphql
```diff
@@ -11,12 +11,12 @@ type TokenHolder @entity {
 	contract: Bytes!
 	address: Bytes!
 	balance: BigInt!
-	allowances: [Allowance!] @derivedFrom(field: "owner")
 }
 
 type Allowance @entity {
 	id: ID!
-	owner: TokenHolder!
+	contract: Bytes!
```
orenyodfat: I guess 'contract' is the 'token' address . right ?
how about rename that to 'token' ?
@ben-kaufman

PR #112, opened 2019-02-18
Fix `npm run migrate`

dOrgJelli:
Repro:
`docker-compose up ganache`
`npm run migrate` (fails)
`npm run codegen` (requires migrate to succeed)

This was introduced here (deployDaoStack.js was deleted):
https://github.com/daostack/subgraph/commit/06f5745e089cde65fba6aad36b108a99645c132f#diff-e847897826ceb8346eb5141f8c23436a

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,4 +71,4 @@ In order to add support for a new contract follow these steps:
          map of solidity event signatures to event handlers in mapping code.
    4. `test/integration/<contract name>.spec.ts`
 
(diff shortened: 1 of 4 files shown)


ben-kaufman:
Thanks! Looks good.
Please just note that if you are using the docker-compose we have here then the DAOstack contracts would be already migrated in the Ganache.

dOrgJelli:
Most definitely, thanks @ben-kaufman!

PR #115, opened 2019-02-19
Add ipfs

### LINKED ISSUE #42: read proposal data from IPFS
# Events 

When creating a new Proposal for a ContributionReward, by calling ContributionReward.proposeContributionReward(...) , the function takes an argument called _contributionDescriptionHash. This hash is  not stored directly on the blockchain, but a NewContributionProposal event is emitted that includes a field called _contributionDescription which has the _contributionDescriptionHash as its value. 

(All this is happening in this contract: https://github.com/daostack/arc/blob/94149aac12325cadc9bd6a989f9e5a5b76e21e2c/contracts/universalSchemes/ContributionReward.sol)

# IPFS
The information we store in IPFS is in JSON format. In theory, we can store all kinds of information. In the current iteration of alchemy, we store "title" and "description", but the content of "desription" is expected to be a URL. Better make this explicit:

{
   "title": "Give me some!",
   "url": "https://...", 
   "description": "A possible very very long description"
}

This would hash to QmT7a1cmrmEY4DN4XPcCuSv8otq6kbudshq5jMWv32XP8R in IPFS.

ben-kaufman:
Fix #42


src/utils.ts
```diff
@@ -3,12 +3,11 @@ import {
   BigInt,
   ByteArray,
   crypto,
-  Entity,
   EthereumEvent,
   store,
   Value,
 } from '@graphprotocol/graph-ts';
-import {Debug } from './types/schema';
+import { Debug } from './types/schema';
```
orenyodfat: is this needed ?

ben-kaufman: The space? No, it just annoyed me 🙃

orenyodfat: The Debug

ben-kaufman: Not exactly necessary, but it's good to keep a debug function as it's sometimes helpful.

PR #119, opened 2019-02-21
New deploy

ben-kaufman:
Support deployment to The Graph servers


orenyodfat:
What this pr is doing ?
produce a migration.json for a single network ?

ben-kaufman:
It's meant to allow a smooth deployment process of the subgraph, both to local and public networks.
I should probably also write that in the README I guess, will add that now.

PR #120, opened 2019-02-24
Add DAO to vote and stake

### LINKED ISSUE #65: proposalVote and ProposlStake are not filterable by DAO
We will (not immediately but at a certain point) need all proposalVotes associated with a particular DAO. 

In other words, `proposalVotes` could use an option to filter by dao address...

### LINKED ISSUE #66: proposalStakes are not filterable by DAO
We will need this at some point. Issue #65  is directly related to this one

ben-kaufman:
Fix #65 
Fix #66

### DIFF
--- a/src/domain/stake.ts
+++ b/src/domain/stake.ts
@@ -19,6 +19,7 @@ export function insertStake(
   staker: Address,
   amount: BigInt,
   proposalId: string,
+  daoId: string,
   outcome: string,
 ): void {
   let stake = getStake(eventId);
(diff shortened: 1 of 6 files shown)


src/domain/stake.ts
```diff
@@ -19,13 +19,15 @@ export function insertStake(
   staker: Address,
   amount: BigInt,
   proposalId: string,
+  daoId: string,
```
orenyodfat: why it is daoId and not dao ?

ben-kaufman: Same reason the above is proposalId and not proposal. You don't need the object just the ID.

orenyodfat: why not the object ?

orenyodfat:
Does this cannot be achieved by queuring proposalVotes using 'where' with a filter ?
 something like proposalVotes where proposal dao equal to dao Id ?

ben-kaufman:
I'm pretty sure that's not possible.

PR #127, opened 2019-02-27
New docker deployment

ben-kaufman:


README.md
```diff
@@ -12,9 +12,9 @@ DAOstack subgraph for [TheGraph](https://thegraph.com/) project.
 Run the tests in the host container:
 
 ```sh
-docker-compose up graph-node
-npm run test # in a new terminal
-docker-compose down -v
+npm run docker:run
+npm run test
+npm run docker:stop
```
jellegerbrandy: how does that work if you are not starting ipfs and postgres?

docker-compose.yml
```diff
@@ -1,36 +1,34 @@
...
+      - '8000:8000'
+      - '8001:8001'
+      - '8020:8020'
     links:
       - ipfs
       - postgres
       - ganache
     environment:
-      postgres_host: postgres:5432
+      postgres_host: 'postgres:5432'
```
jellegerbrandy: ?? why?

ops/set-docker-network.js
```diff
@@ -0,0 +1,41 @@
...
+  } catch (msg) {
+    throw Error(`Setting docker network failed! ${msg}`)
+  }
+}
+
+if (require.main === module) {
+  setDockerNetwork().catch((err) => { console.log(err); process.exit(1) })
+} else {
+  module.exports = setDockerNetwork
+}
```
jellegerbrandy: I personally would have preferred separate docker-compose files, which is more explicit then  this magical  rewriting of the file (which version fo docker-compose are you going to commit now?).  But hey, whatever works :-)

jellegerbrandy: Also, can you update the documentation and explain what the .env file should look like?

PR #129, opened 2019-02-28
release.sh script to publish docker images

jellegerbrandy:


release.sh
```diff
@@ -0,0 +1,36 @@
...
+migration_version=$(cat package.json  | jq -r '.devDependencies."@daostack/migration"')
+package_version=$(cat package.json | jq -r '.version')
+image_version=ganache-$migration_version-$package_version
+echo "docker commit $container_id $image_name:$image_version"
+docker commit $container_id $image_name:$image_version
+echo "docker push $image_name:$image_version"
+docker push $image_name:$image_version
+
+docker-compose down -v
+# publish npm
```
ben-kaufman: Please uncomment or remove

jellegerbrandy: Ah, yes, the subgraph name fromt he .env file is no good. This is for a ganache deployment, and the name can just be `daostack`. I'll work around that.

PR #140, opened 2019-03-10
Add npm package release to release script

ben-kaufman:


jellegerbrandy:
I am not against this, but we have not making any releases on npmjs up to now, and i am not sure if we will every need them.
(Currently, there is a dependency on @doastack/subgraph in the package.json of client and alchemy, but I think we can remove these dependencies (and just use the docker images))

ben-kaufman:
I just uploaded to NPM yesterday as a developer wanted to us that, so I'm not sure we need it but maybe others will find it useful.

jellegerbrandy:
Did you get feedback from "a developer"? it does not have a `main` section in the package, so it will not be importable, I think. And also, there is no code in this package, as far as I can see, that you'd want to import.

ben-kaufman:
I'm not sure exactly what he needed from that but I think it's still good to have just in case.

ben-kaufman:
This would be useful now as we use subgraph as an NPM package. @orenyodfat @jellegerbrandy please take a look.

PR #151, opened 2019-03-17
Fix #147

### LINKED ISSUE #147: gpRewards contains records that have no claimable rewards at all
If I run this query ```{ gprewards {
    id
  
  	tokensForStaker
  reputationForVoter
  reputationForProposer
  daoBountyForStaker
  
    
  }
} ```

I see records such as:

```
      {
        "daoBountyForStaker": null,
        "id": "0xd338579ef3105083c155df9a92e27bf77d754dd8c58d455bad9ed3f00f043b2c",
        "reputationForProposer": null,
        "reputationForVoter": null,
        "tokensForStaker": null
      }
  
```
These are not useful in any way, as far as I can see. Am I missing a point, or is this a bug?

ben-kaufman:


test/integration/domain.spec.ts
```diff
@@ -811,6 +811,13 @@ describe('Domain Layer', () => {
             accountsWithUnclaimedRewards
         }
     }`;
+
+    const proposalIsIndexed = async () => {
+      return (await sendQuery(getProposalRewards)).proposal.accountsWithUnclaimedRewards.length === 2;
+    };
```
orenyodfat: How about adding a test to check the actual proposal.accountsWithUnclaimedRewards ?

orenyodfat: for proposer.voters and stakers

ben-kaufman: There is a test for that already...

PR #153, opened 2019-03-20
Set DAO addresses

### LINKED ISSUE #152: The dao collection should only return indexed DAOs
At the moment, the daos colection contains all registered daos, but they are not all being indexed by the subgraph.

We need a way to distinguish them, so we can show only the working DAO instances in alchemy. 

This can be solved in one of two ways:

- only return DAO entitities that are indexed
- add a `isIndexed` property to  the DAO entity

I personally prefer the first.

Related front-end issue: https://github.com/daostack/alchemy/issues/437#event-2208603649

ben-kaufman:
Fix #152

### DIFF
--- a/ops/generate-subgraph.js
+++ b/ops/generate-subgraph.js
@@ -53,12 +53,13 @@ async function generateSubgraph () {
     dataSources
   }
 
-  console.log('Use this addresses list in utils isAddressIndexed function:\n')
-  let indexedArrayStirng = ''
+  let indexedArrayStirng = 'import { Address } from \'@graphprotocol/graph-ts\';\n\nexport let addresses: Address[] = [\n'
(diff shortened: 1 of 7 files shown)


ben-kaufman:
@jellegerbrandy This should be working well.
The main issue I have now is with testing the UController. I currently removed its tests but I don't think it's a good idea... Anyway, I still don't have an idea on how to do these testings properly now, so if you have an idea please let me know.

orenyodfat:
What this pr is trying to solve ?

jellegerbrandy:
Oren: This PR is trying to solve #152, as it says in the description.

Ben: What is the problem with testing the UController?

orenyodfat:
Issue 152 should be solved by the arc-hive

PR #162, opened 2019-03-26
support DAORegistry 

orenyodfat:
- doa entity has a new field : `register`

- registered values can be : 
  `na` - no one proposed nor register that.
  `proposed` - it was proposed by someone
  `registered` -it was registered as a legitimate dao (by the owner of the registry).
  `unRegistered` - unregistered by the owner of the registry  - which in a sense same as `na`


src/mappings/DAORegistry/datasource.yaml
```diff
@@ -0,0 +1,9 @@
+entities:
+  - Dummy
```
ben-kaufman: Please delete this

PR #173, opened 2019-04-07
Make GPQueue and Controller Scheme useful!

### LINKED ISSUE #171: Make controllerschemes and gpQueues more usable
opened on behalf of @jellegerbrandy 
The new `gpQueue`, `ucontrollerScheme`, `controllerScheme` and `ControllerOrga
-nization`
are all free-standing entities, not related to the other graph entities at all.

This makes them difficult and expensive to use.

In particular, what we need is:

* `Proposal.gpQueue`: the `gpQueue` that the proposal is part of **<- plz make this first**
* `DAO.controllerSchemes`
* `DAO.uControllerSchemes`
* (given that the DAO either uses controller and ucontroller (and never both, is that correct?), it would simplify the data model is there is just a singe "controllerScheme" entity
instead of a controllerScheme and a uControllerScheme with exactly the same data model, so we can have just one "controllerScheme" entity that
contains data either entity...)
* `DAO.gpQueues`: the gpQueues of these DAO

ben-kaufman:
Fix #171

### DIFF
--- a/src/domain/proposal.ts
+++ b/src/domain/proposal.ts
@@ -1,7 +1,7 @@
 import { Address, BigInt, Bytes, ipfs, json, JSONValueKind, store } from '@graphprotocol/graph-ts';
 import { GenesisProtocol } from '../types/GenesisProtocol/GenesisProtocol';
 import { Proposal } from '../types/schema';
-import { equals, equalsBytes } from '../utils';
+import { bytesFromString, equals, equalsBytes } from '../utils';
 import { updateThreshold } from './gpqueue';
 
 export function parseOutcome(num: BigInt): string {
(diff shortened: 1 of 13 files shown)


src/domain/proposal.ts
```diff
@@ -134,6 +134,13 @@ export function updateGPProposal(
   proposal.organizationId = gpProposal.value0;
   proposal.expiresInQueueAt = timestamp.plus(params.value1);
   proposal.createdAt = timestamp;
+  updateThreshold(
+    proposal.dao.toString(),
+    gp.threshold(proposal.paramsHash, proposal.organizationId),
```
orenyodfat: proposal.organizationId seems to be redundant. as it is the id of GPQUE.

test/integration/SchemeRegistrar.spec.ts
```diff
@@ -171,7 +171,7 @@ describe('SchemeRegistrar', () => {
 
         const executedIsIndexed = async () => {
           return (await sendQuery(getSchemeRegistrarProposalExecuteds)).schemeRegistrarProposalExecuteds.length
-           > prevExecutedsLength;
+           > prevExecutedsLength + 1;
```
orenyodfat: why and how this pr effect getSchemeRegistrarProposalExecuteds length ?

ben-kaufman: That's a bug I fixed is not exactly related.

src/domain/proposal.ts
ben-kaufman: It causes problems to remove it and Alchemy might be using it.

PR #177, opened 2019-04-08
try fo fix the docker images

jellegerbrandy:
The generated docker images break all the up-stream tests. There are two problems (a) the test dao is not added to the indexed cocntracts and (b) the graph-node starts up, but only indexes block 0

- for fix (a), I added a dao in `daos/private`. I copied the data from `migration.json` by hand, This should probably be done automatically in the `release.sh` script
- (b) was not a subgraph problem :-)

### DIFF
--- a/daos/private/test.json
+++ b/daos/private/test.json
@@ -1,0 +1,7 @@
+{
+    "name": "Uxorious Model",
+    "Avatar": "0xE7A2C59e134ee81D4035Ae6DB2254f79308e334f",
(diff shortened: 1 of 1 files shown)


orenyodfat:
There is already a code which automatically set the test dao
https://github.com/daostack/subgraph/blob/master/ops/generate-daos-subgraph.js#L34 
Not sure what this pr is trying to solve.  @jellegerbrandy

PR #184, opened 2019-04-14
Added scheme property to proposal entity.

### LINKED ISSUE #180: add more fields on `gpQueue` object

### LINKED ISSUE #182: add "scheme" propoerty to proposal entity
at this moment, this should be one of `GenericScheme`, `ContributionReward` or `SchemeRegistrar`

### LINKED ISSUE #187: Support proposal IPFS descriptions for all proposal schemes
Currently it is only for CR
- consider add that at getProposal on the creation of new Proposal
- consider replace ipfs.cat with `ipfs.mapJSON(hash: string, callback: string, userData: Value) for files with JSON values `

ben-kaufman:
Fix #182 

partially fix #180 

fix #187


docker-compose.yml
```diff
@@ -1,8 +1,7 @@
 version: '3'
 services:
   graph-node:
-    image: >-
-      graphprotocol/graph-node:v0.6.0
+    image: 'graphprotocol/graph-node:v0.6.0'
```
orenyodfat: why this ?

orenyodfat:
please add tests

src/domain/gpqueue.ts
```diff
@@ -1,21 +1,52 @@
...
+import { Address, BigInt, Bytes } from '@graphprotocol/graph-ts';
+import { GenesisProtocol } from '../types/GenesisProtocol/GenesisProtocol';
 import { GPQueue } from '../types/schema';
+import { debug } from '../utils';
 
 export function getGPQueue(id: string): GPQueue {
   let gpQueue = GPQueue.load(id) ;
   if (gpQueue == null) {
     gpQueue = new GPQueue(id);
+    gpQueue.votingMachine = null;
```
orenyodfat: is this a must ?

src/domain/gpqueue.ts
```diff
@@ -1,21 +1,52 @@
...
+    gpQueue.minimumDaoBounty = params.value9; // minimumDaoBounty
+    gpQueue.daoBountyConst = params.value10; // daoBountyConst
+    gpQueue.activationTime = params.value11; // activationTime
+    gpQueue.voteOnBehalf = params.value12; // voteOnBehalf
+  }
   gpQueue.dao = dao;
   gpQueue.save();
 }
+
+export function setScheme(
```
orenyodfat: why gpque need to have the scheme name ?
do we need this function?
why scheme cannot be passed as a param to updateThreshold ?

src/domain/proposal.ts
```diff
@@ -186,12 +208,38 @@ export function updateGSProposal(
...
+  schemeAddress: Address,
 ): void {
   let proposal = getProposal(proposalId.toHex());
   proposal.dao = avatarAddress.toHex();
   proposal.genericScheme = proposalId.toHex();
   proposal.createdAt = createdAt;
   proposal.descriptionHash = descriptionHash;
+  proposal.scheme = crypto.keccak256(concat(avatarAddress, schemeAddress)).toHex();
+  if (proposal.organizationId != null && proposal.scheme != null) {
+    setScheme(proposal.organizationId, proposal.scheme);
```
orenyodfat: why setting scheme at gpque on each proposal ?

src/domain/proposal.ts
```diff
@@ -186,12 +208,38 @@ export function updateGSProposal(
...
+    setSchemeName(proposal.scheme, 'GenericScheme');
+  }
+  // IPFS reading
+
+  let ipfsData = ipfs.cat('/ipfs/' + descriptionHash);
```
orenyodfat: this IPFS code is common to all schemes .
lets have it on a function .how about doing the IPFS staff of the getProposal function.

.gitignore
```diff
@@ -7,3 +7,4 @@ yarn.lock
 src/types
 /schema.graphql
 .env
+build/
```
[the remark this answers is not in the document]
ben-kaufman: I'm not sure... The schema.graphql is created in the ops scripts

src/domain/gpqueue.ts
ben-kaufman: Yes, it will crash without this

src/domain/proposal.ts
ben-kaufman: We can add a check if that's already set...

src/domain/proposal.ts
ben-kaufman: ok

src/domain/gpqueue.ts
ben-kaufman: Because you don't necessarily have the scheme whenever calling updateThreshold

orenyodfat: for example when ?

src/domain/proposal.ts
orenyodfat: not sure it is the right place here.

src/domain/gpqueue.ts
ben-kaufman: https://github.com/daostack/subgraph/pull/184/files#diff-b75c7488932d0055e4cddbd7f02c2ed7R179

orenyodfat: well , in that case you can get the scheme from the proposal it self  as you already reading that at https://github.com/daostack/subgraph/pull/184/files#diff-b75c7488932d0055e4cddbd7f02c2ed7R170 . proposal.callbacks (value01).

PR #193, opened 2019-04-17
add dao isExist function

### LINKED ISSUE #192: index only proposals on an indexed dao

orenyodfat:
fix #192

### DIFF
--- a/src/domain/dao.ts
+++ b/src/domain/dao.ts
@@ -58,3 +58,9 @@ export function register(
     dao.save();
   }
 }
+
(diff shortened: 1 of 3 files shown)


src/domain/dao.ts
```diff
@@ -58,3 +58,9 @@ export function register(
     dao.save();
   }
 }
+
+export function isExist(
```
jellegerbrandy: this is not English. Either `doesExist` or `exists`

PR #201, opened 2019-04-23
Improve QEP handling

ben-kaufman:


src/domain/proposal.ts
```diff
@@ -77,9 +77,16 @@ export function updateProposal(
 ): void {
   let gp = GenesisProtocol.bind(gpAddress);
   let gpProposal = gp.proposals(proposalId);
+  let prevOutcome = proposal.winningOutcome;
   proposal.votingMachine = gpAddress;
   // proposal.winningVote
   proposal.winningOutcome = parseOutcome(gpProposal.value3);
+  if ((equalStrings(proposal.stage, 'Boosted') || equalStrings(proposal.stage, 'QuietEndingPeriod'))
+   && !equalStrings(proposal.winningOutcome, prevOutcome)) {
+    if (gpProposal.value2 === 6) {
```
orenyodfat: Why not just log the state ? For any case

ben-kaufman: There is test in the client for that

orenyodfat:
We need a test inside this repo . Like all other tests .

PR #210, opened 2019-04-30
Fix unclaimed redeemers list

### LINKED ISSUE #206: accountsWithUnclaimedRewards removes addresses also when rewards are still unclaimed
If we redeem, for example, only the nativeToken rewards but not the ETH rewards, the address of the beneificiary will be removed from accountsWithUnclaimedRewards. But the ETH rewards are still there to redeem.

This occurs naturally if we use the `Redeem.contributionRewardRedeem` function when the ETH balance of the DAO avatar is not high enough to reward the user.

ben-kaufman:
Fix #206

### DIFF
--- a/src/domain/proposal.ts
+++ b/src/domain/proposal.ts
@@ -2,7 +2,7 @@ import { Address, BigInt, Bytes, crypto, ipfs, json, JSONValueKind, store } from
 import { setSchemeName } from '../mappings/Controller/mapping';
 import { GenesisProtocol } from '../types/GenesisProtocol/GenesisProtocol';
 import { Proposal } from '../types/schema';
-import { concat, equals, equalsBytes, equalStrings, debug } from '../utils';
+import { concat, equals, equalsBytes, equalStrings } from '../utils';
 import { updateThreshold } from './gpqueue';
 
 export function parseOutcome(num: BigInt): string {
--- a/src/domain/proposal.ts
+++ b/src/domain/proposal.ts
@@ -204,10 +204,9 @@ export function updateCRProposal(
   proposal.scheme = crypto.keccak256(concat(avatarAddress, schemeAddress)).toHex();
   setSchemeName(proposal.scheme, 'ContributionReward');
   getProposalIPFSData(proposal);
-
+  addRedeemableRewardOwner(proposal, beneficiary);
   saveProposal(proposal);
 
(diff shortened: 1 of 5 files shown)


src/domain/proposal.ts
```diff
@@ -204,6 +205,8 @@ export function updateCRProposal(
   setSchemeName(proposal.scheme, 'ContributionReward');
   getProposalIPFSData(proposal);
 
+  addRedeemableRewardOwner(proposalId, beneficiary);
```
orenyodfat: why is this needed ?

src/domain/reward.ts
```diff
@@ -81,6 +81,32 @@ export function reputationRedemption(proposalId: Bytes, beneficiary: Address, ti
 }
 
 function shouldRemoveAccountFromUnclaimed(reward: GPReward): boolean {
+  let proposal = getProposal(reward.proposal);
+  if (proposal.contributionReward !== null) {
+    let crProposal = store.get(
```
orenyodfat: please use ContributionRewardProposal.load

src/domain/reward.ts
```diff
@@ -81,6 +81,32 @@ export function reputationRedemption(proposalId: Bytes, beneficiary: Address, ti
 }
 
 function shouldRemoveAccountFromUnclaimed(reward: GPReward): boolean {
+  let proposal = getProposal(reward.proposal);
```
orenyodfat: why need this ? 
use ContributionRewardProposal.load(pid) and check if its not null.

orenyodfat:
please add tests to cover this issue.

src/domain/proposal.ts
ben-kaufman: To make sure the beneficiary is listed as a redeemer

src/domain/reward.ts
```diff
@@ -81,6 +81,28 @@ export function reputationRedemption(proposalId: Bytes, beneficiary: Address, ti
 }
 
 function shouldRemoveAccountFromUnclaimed(reward: GPReward): boolean {
+  let proposal = ContributionRewardProposal.load(reward.proposal);
+  if (proposal !== null) {
+    if (equalsBytes(proposal.beneficiary, reward.beneficiary)) {
```
orenyodfat: does this cover by any test?

PR #214, opened 2019-05-02
Fix AvatarContract saving

### LINKED ISSUE #213: GraphQL error: Null value resolved for non-null field `address`
running the following query:
 ```
{
            avatarContracts {
              address
            }
          }
```

ben-kaufman:
Fix #213

### DIFF
--- a/src/mappings/Avatar/mapping.ts
+++ b/src/mappings/Avatar/mapping.ts
@@ -37,7 +37,13 @@ export function handleReceiveEth(event: ReceiveEther): void {
 export function handleOwnershipTransferred(event: OwnershipTransferred): void {
   let avatar = AvatarContract.load(event.address.toHex());
   if (avatar == null) {
-      avatar = new AvatarContract(event.address.toHex());
+    avatar = new AvatarContract(event.address.toHex());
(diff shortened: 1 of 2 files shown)


orenyodfat:
please add test to verify that

ben-kaufman:
I'm not sure it's possible to test this scenario, this happens only for a moment and we already have a test for `avatarContracts`.

orenyodfat:
What happend in this moment ?

ben-kaufman:
I'm pretty sure it's before the first scheme is registered but the Avatar ownership was already transferred. I think I can add this check to the start of the domain test but this would be pretty strange for the test as this belongs to the Avatar and not domain testing.

orenyodfat:
the avatar ownership will be transferred before the first scheme is registered most of the times.

ben-kaufman:
Ok, but how would you suggest to write a good test for that?

orenyodfat:
how this issue happened?

ben-kaufman:
It happened because the code to save the Avatar was present twice, but was saving the full data needed only once. I just changed it to do save it a single time.

Issue #208, opened 2019-04-30
members collection is populated with addreses that have 0 rep and 0 tokens


jellegerbrandy:
Got here playing around with alchemy and passing around rewards, so i do not have precise steps to reproduce. I find this in the collection.

What I expect is that all members either have a > 0 tokens or > 0 reputation.

![image](https://user-images.githubusercontent.com/1306173/56957074-47eb5a80-6b46-11e9-8da3-a38a69b82a39.png)
![image](https://user-images.githubusercontent.com/1306173/56957105-5cc7ee00-6b46-11e9-8b40-493536e8ebda.png)

jellegerbrandy:
Actually, the last two records are also in the v16 docker images. So it should be easy to reproduce :-)

ben-kaufman:
This is no bug, this is the Avatar. Though this part does need some small refactoring but it's not related.
You requested in the past to have the Avatar as a member, I can remove that but I guess you do need it.

jellegerbrandy:
Okay, we are indeed using that. Sorry for the confusion.
And the first record with the NULL address?

tibetsprague:
Wait, why do we want to show the avatar as a "member", thats confusing.

ben-kaufman:
@jellegerbrandy couldn't reproduce that 0 address, can you give some details on how you created that?

PR #216, opened 2019-05-05
Refactor members adding/ removing.

ben-kaufman:
Small refactoring.

### DIFF
--- a/src/domain/member.ts
+++ b/src/domain/member.ts
@@ -13,6 +13,7 @@ export function getMember(address: Address, daoAddress: Address): Member {
     member.address = address;
     member.dao = daoAddress.toHex();
     member.reputation = BigInt.fromI32(0);
+    member.tokens = BigInt.fromI32(0);
     saveMember(member);
     increaseDAOmembersCount(member.dao);
   }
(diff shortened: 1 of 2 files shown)


orenyodfat:
is this related to specific issue ?

ben-kaufman:
Not exactly. I did it when investigating this issue #208 but it's not exactly related (probably).

src/domain/member.ts
```diff
@@ -50,7 +51,7 @@ export function updateMemberReputationWithValue(
 ): void {
   let member = getMember(address, daoAddress);
   member.reputation = value;
-  if (equals(member.reputation as BigInt, BigInt.fromI32(0))) {
+  if (equals(member.reputation, BigInt.fromI32(0)) && equals(member.tokens, BigInt.fromI32(0))) {
```
orenyodfat: does and why token holders are dao members ?

ben-kaufman: Yes and because that's what Alchemy needed

orenyodfat: plus @jellegerbrandy for clarification here.

jellegerbrandy: No, I think it is clearer if we members collection only contanis the reputation holders. (So I'd prefer not to merge this change)

PR #218, opened 2019-05-13
Add gpQueue to controllerScheme

### LINKED ISSUE #217: add a scheme --> gpQueue edge
We already have `gpQueue.scheme`;  what we really need is the inverse `controllerscheme.queue` edge.

ben-kaufman:
Fix #217

### DIFF
--- a/src/domain/gpqueue.ts
+++ b/src/domain/gpqueue.ts
@@ -16,7 +16,6 @@ export function getGPQueue(id: string): GPQueue {
 export function updateThreshold(dao: string,
                                 gpAddress: Address,
                                 threshold: BigInt,
-                                paramsHash: Bytes,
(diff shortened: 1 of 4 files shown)


orenyodfat:
Why adding gpq yo controller scheme ?
Not all controller schemes as gpq

jellegerbrandy:
The reason is this. The client lib (wants to) implement the following pattern to get the queues:

- get all the schemes
- those schemes that have a gpQueue in the subgraph are returned 
- those schemes that do not have a gpQueue in the subgraph: if these are schemes of which we know that they have a gpQueue (i.e. schemes that do not have a first proposal yet but we know from migration.json which they are), also those are returned
- the other schemes are ignored

to implement that in a slightly efficient way, it's this kind of query we need....

PR #228, opened 2019-05-16
remove domain member entity

### LINKED ISSUE #208: members collection is populated with addreses that have 0 rep and 0 tokens

orenyodfat:
- member entity was removed
- using reputationHolders instead
- avatar is not a reputationHolders 
- fix #208 

please note that a member entity was holds the dao token balance of a user. now it is gone.


src/mappings/Reputation/mapping.ts
```diff
@@ -36,12 +30,34 @@ function update(contract: Address, owner: Address): void {
...
-    reputationHolders.push(ent.id);
+  let repHolderId = crypto.keccak256(concat(contract, owner)).toHex();
+  let repHolder = ReputationHolder.load(repHolderId);
+  let balance = rep.balanceOf(owner);
+  if (repHolder == null) {
+    repHolder = new ReputationHolder(repHolderId);
+    repHolder.contract = contract;
+    repHolder.address = owner;
+    repHolder.balance = balance;
+    if (!equals(balance, BigInt.fromI32(0))) {
```
jellegerbrandy: so you are also creating a repHolder here if the balance is 0?

orenyodfat:
Only if the balance is not 0

Issue #222, opened 2019-05-14
contract versions 


orenyodfat:
- [x] Subgraph will test all versions
- [x] Subgraph will include latest migration
- [x] Add arc versions to ops/mapping.json
- [x] Update generate-abi script to get abi via migration package
- [x] Update generate-subgraph script to use abi per version.
- [ ] Having contract versions mapping - (address:version) -low priority

depend on  https://github.com/daostack/migration/issues/130

PR #229, opened 2019-05-16
Multiple Arc versions support.

ben-kaufman:
Solves #222


.travis.yml
```diff
@@ -34,4 +34,5 @@ script:
   - echo -en 'travis_fold:end:script.2\\r'
 
   - npm run lint
+  - sleep 120
```
orenyodfat: lets remove that and add subscription to subgraph to check if it is already synced .
we should add this check for each test and before each of one.

orenyodfat:
please update migration to include https://github.com/daostack/migration/pull/135

ben-kaufman:
It would be much better as you describe but the issue is that different versions may have different behaviors, so the tests might be different for each version.

PR #236, opened 2019-05-22
add ContractInfo

orenyodfat:
this will index contracts addresses ,name and its version


jellegerbrandy:
Now that we ahve the contact names availabe in the subgraph itself, we can populate the `scheme.name` fields immediately and do not have to wait for proposal creation. Right?

orenyodfat:
we might .yes

> Now that we ahve the contact names availabe in the subgraph itself, we can populate the `scheme.name` fields immediately and do not have to wait for proposal creation. Right?

we might .yes

PR #244, opened 2019-05-27
gpparams at schemes

### LINKED ISSUE #238: populate scheme.name
Now that we ahve the contact names availabe in the subgraph itself, we can populate the scheme.name fields immediately and do not have to wait for proposal creation.

### LINKED ISSUE #239: add governanceTimeStart to schemes

### LINKED ISSUE #240: we need a way to get the target contract of a genericScheme from the scheme itself
What we'd like to do in Alchemy: show the address of the contract that will be called before actually creating the proposal - for example, to show the dutchX forms, we need to know that this GenericScheme is calling the dutchx contract.

As far asI can see, this info is not available in the subgraph at the moment.

I am not sure what the best place to store it is....

orenyodfat:
fix #239 
fix #240 
fix #238

### DIFF
--- a/daos/private/testdao.json
+++ b/daos/private/testdao.json
@@ -1,5 +1,5 @@
 {
-  "name": "Bilious Toilet",
+  "name": "Comely Engineering",
   "Avatar": "0x3b8847C03D86FF38E79f915b23f6531DF3bA01e1",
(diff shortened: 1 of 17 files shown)


dkent600:
Fixed how?  What are the API changes here?

PR #249, opened 2019-05-30
Fix bug with GP parameters

ben-kaufman:


ben-kaufman:
This issue should be further investigated, but I can't do that right now and as far as I understood this release is blocking issue for the Alchemy dev team so only merge this if really it's urgent.

orenyodfat:
Does the issue reproduced ?
Is there a test to verify it is fixed ?

src/mappings/Controller/mapping.ts
```diff
@@ -268,7 +268,7 @@ export function setGPParams(gpAddress: Address,
                             gpParamsHash: Bytes): void {
     let gp = GenesisProtocol.bind(gpAddress);
     let gpParams = GenesisProtocolParam.load(gpParamsHash.toHex());
-    if (gpParams == null) {
+    if (gpParams == null && !equalsBytes(gpParamsHash, new Bytes(32))) {
```
orenyodfat: yes, there can be a case where someone register a scheme with gp param as 0.by mistake or purpose .

PR #253, opened 2019-06-03
check if scheme is connected to the dao upon calculate its redeem values.

### LINKED ISSUE #248: set IPFS timeout to 2 seconds + 100 kb file size limit
https://github.com/graphprotocol/graph-node/blob/master/docs/environment-variables.md#running-mapping-handlers

### LINKED ISSUE #250: v19 (and v20) fail on rinkeby
Basically, they stop indexing proposals.

this is the error message:
```
ubuntu@ip-172-31-22-45:~/subgraph-rinkeby$ sudo docker-compose logs graph-node | grep ERR
graph-node_1  |      May 31 14:27:18.646 ERRO slog-async: logger dropped messages due to channel overflow, count: 127
graph-node_1  |    May 31 14:47:31.497 ERRO Subgraph instance failed to run: Failed to process trigger: Failed to handle Ethereum event with handler "handleStateChange": Trap: Trap { kind: Host(HostExportError("Failed to call function \"redeem\" of contract \"GenesisProtocol\": ABI error: please ensure the contract and method you\'re calling exist! failed to decode empty bytes. if you\'re using jsonrpc this is likely due to jsonrpc returning `0x` in case contract or method don\'t exist")) }, code: SubgraphSyncingFailure, id: QmPEKgbFN2a5NsiYG8FVzV4NksebT21Dud5MdrVBtx89VJ
```
And this is the empty rpoposals list: 

https://rinkeby.subgraph.daostack.io/subgraphs/name/v19/graphql?query=%7B%0A%20proposals%20%7B%0Atitle%0A%7D%0A%7D

orenyodfat:
fix #250

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,7 +16,10 @@ services:
       postgres_pass: letmein
       postgres_db: postgres
       ipfs: 'ipfs:5001'
-      ethereum: 'development:http://ganache:8545'
+      ethereum: 'rinkeby:https://rinkeby.infura.io/v3/69b4a42f71a8446e8b167062e3d5e317'
+      GRAPH_IPFS_TIMEOUT: '2'
+      GRAPH_MAX_IPFS_FILE_BYTES: '102400'
+      RUST_LOG: info
   ipfs:
     build: docker/ipfs
     ports:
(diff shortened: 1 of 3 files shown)


ben-kaufman:
This indeed solves the bug which is great, but the limit on the IPFS time causes the ipfs data not to be pulled (at least on my machine). I'll remove that from here if it's okay and we can open a dedicated PR for that.

PR #258, opened 2019-06-04
ipfs settings 

### LINKED ISSUE #248: set IPFS timeout to 2 seconds + 100 kb file size limit

orenyodfat:
fix #248

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -17,7 +17,8 @@ services:
       postgres_db: postgres
       ipfs: 'ipfs:5001'
       ethereum: 'development:http://ganache:8545'
-      RUST_LOG: info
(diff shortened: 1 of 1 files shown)


ben-kaufman:
I coudn't successfully retrieve IPFS data with those parameters. I guess that's because the IPFS instance on my machine wasn't synced with the relevant data. This change means we will need to make changes on how we setup a new IPFS node as we will have to make sure we have all necessary data on the IPFS before running the GraphNode, also we will have to write directly to the IPFS instance we use here from Alchemy (if we don't already).

orenyodfat:
me and @jellegerbrandy confirmed it is ok.
seems like you encounter other issues

PR #266, opened 2019-06-23
Added totalRepWhenCreated to proposal entity

### LINKED ISSUE #265: add the total reputation supply at the moment of proposal creation
as the reputation is "freeze" at the event of proposal creation .
this value should be used to calculate voting percentages.

ben-kaufman:
Fix #265

### DIFF
--- a/src/domain/proposal.ts
+++ b/src/domain/proposal.ts
@@ -2,7 +2,9 @@ import { Address, BigInt, Bytes, crypto, ipfs, json, JSONValueKind, store } from
 import { GenesisProtocol } from '../types/GenesisProtocol/GenesisProtocol';
 import { ControllerScheme, Proposal } from '../types/schema';
 import { concat, equalsBytes, equalStrings } from '../utils';
+import { getDAO } from './dao';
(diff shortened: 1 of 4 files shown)


src/domain/proposal.ts
```diff
@@ -172,6 +174,10 @@ export function updateGPProposal(
     scheme.gpQueue = proposal.organizationId.toHex();
     scheme.save();
   }
+
+  let dao = getDAO(avatarAddress.toHex());
```
orenyodfat: this should be added to all proposals (not just one type)
how about adding that to the getProposal function

test/0.0.1-rc.19/domain.spec.ts
```diff
@@ -495,6 +496,7 @@ describe('Domain Layer', () => {
       quietEndingPeriodBeganAt: null,
       executedAt: null,
       totalRepWhenExecuted: null,
+      totalRepWhenCreated: totalRep,
```
orenyodfat: you should read the totalRep just after the proposal creation

PR #291, opened 2019-07-17
Fix members count bug

ben-kaufman:
Fixed rep holders count.
Fixed mainnet mappings issue

### DIFF
--- a/daos/mainnet/testdao.json
+++ b/daos/mainnet/testdao.json
@@ -1,8 +1,0 @@
-{
-  "name": "Spasmodic Chair",
-  "Avatar": "0xd05e2aB61fEd443BFF835B9850d50205ACa4f1Fa",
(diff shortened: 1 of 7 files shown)


daos/mainnet/testdao.json
```diff
@@ -1,7 +1,8 @@
 {
```
orenyodfat: why this file changed ?
please add it to .gitignore

ops/mappings.json
```diff
@@ -591,20 +591,6 @@
       "mapping": "Redeemer",
       "arcVersion": "0.0.1-rc.16"
     },
-    {
```
orenyodfat: how this is related ?

ops/mappings.json
```diff
@@ -730,6 +688,13 @@
       "dao": "base",
       "mapping": "Redeemer",
       "arcVersion": "0.0.1-rc.21"
+    },
+    {
```
orenyodfat: how this is related ?

PR #292, opened 2019-07-18
Update closingAt on execution

### LINKED ISSUE #289: closingAt does not work as expeced
the field would be much more useful for us if "closingAt" was set to the current time at the moment of execution.

see for a description of the problem here: https://github.com/daostack/alchemy/issues/901

ben-kaufman:
Fix #289


orenyodfat:
please add test .

PR #303, opened 2019-08-08
make subgraph usable as a package, and add some parametrization

jellegerbrandy:
this PR should not change any functionality, just add the possiblity to require the package and use it as a dependency (which is useful for the test_env repo) and add some parameters (i.e. tell the repo to use  a specific migraiton.json, not the one copied in the package itself)

### DIFF
--- a/ops/generate-daos-subgraph.js
+++ b/ops/generate-daos-subgraph.js
@@ -5,16 +5,6 @@ const daodir = "./daos/" + network + "/";
 const path = require("path");
 const currentDir = path.resolve(`${__dirname}`)
 
-/**
- * ?????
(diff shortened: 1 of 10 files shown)


ops/generate-daos-subgraph.js
```diff
@@ -1,11 +1,23 @@
 const fs = require("fs");
 const yaml = require("js-yaml");
-const { migrationFileLocation, network } = require("./settings");
+const { migrationFileLocation: defaultMigrationFileLocation, network } = require("./settings");
 const daodir = "./daos/" + network + "/";
+const path = require("path");
+const currentDir = path.resolve(`${__dirname}`)
 
+/**
+ * ?????
```
ben-kaufman: Please remove

PR #312, opened 2019-08-27
Fix deployment issues

ben-kaufman:


ops/graph-deploy.js
```diff
@@ -15,22 +15,21 @@ async function deploy (opts = {}) {
...
-  // if (result[0] === 1) {
-  //   console.error(`Create failed! ${msg}`)
-  // }
-  if (msg.toLowerCase().indexOf('error') > 0) {
-    if (msg.match(/subgraph already exists/)) {
-      // the subgraph was already created before -we're ok
-      console.log('subgraph already exists - deploying a new version')
-    } else {
-      console.error(`Create failed! ${msg}`)
+  if (graphNode !== 'https://api.thegraph.com/deploy/') {
```
orenyodfat: what is the 'else' of this condition ?

ben-kaufman: There should be no else.

orenyodfat: why this is needed ?

ben-kaufman: Because you can't create subgraph which is on The Graph's hosted services, so only if it's not that you should create it.

orenyodfat: oh, ok ..need to be able to do that.

PR #313, opened 2019-08-28
Change IPFS limit

ben-kaufman:
Fix IPFS limit to 2700 bytes.


orenyodfat:
please use graph node latest which include https://github.com/graphprotocol/graph-node/pull/1138 .

orenyodfat:
this issue is handled by #314

PR #334, opened 2019-09-22
Add proposal counting per schema

### LINKED ISSUE #326: keep counts of proposals
We are doing a lot of separate queries to count proposals, and we coudl avoid that if these counters are part of the index. So I'd like to add the folllowing fields:

``` 
Scheme.numberOfQueuedProposals
Scheme.numberOfPreBoostedProposals
Scheme.numberOfBoostedProposals
DAO.numberOfQueuedProposals
DAO.numberOfPreBoostedProposals
DAO.numberOfBoostedProposals
```
For the other states (`ExpiredInQueue`, `Executed`, `QuietEndingPeriod`) we are not doing separate queries, but it could be useful to have these counts available anyway to help us with paging and showing totals.

ben-kaufman:
Fix #326

### DIFF
--- a/src/mappings/Controller/schema.graphql
+++ b/src/mappings/Controller/schema.graphql
@@ -23,7 +23,9 @@ type ControllerScheme @entity {
 	schemeRegistrarParams: SchemeRegistrarParam
 	uGenericSchemeParams: UGenericSchemeParam
 	genericSchemeParams: GenericSchemeParam
-
+	numberOfQueuedProposals: BigInt!
+	numberOfPreBoostedProposals: BigInt!
+	numberOfBoostedProposals: BigInt!
 }
 
 type ControllerGlobalConstraint @entity {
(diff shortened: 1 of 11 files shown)


src/mappings/Controller/schema.graphql
```diff
@@ -23,7 +23,9 @@ type ControllerScheme @entity {
 	schemeRegistrarParams: SchemeRegistrarParam
 	uGenericSchemeParams: UGenericSchemeParam
 	genericSchemeParams: GenericSchemeParam
-
+	numberOfQueuedProposals: BigInt!
```
orenyodfat: is that not suppose to be per dao ?

jellegerbrandy: yes, indeed it is, good call.

It is probably better to count and store this ont he `gpqueue` object.

orenyodfat:
Though gpque is per scheme (not dao) as well . Though if this what’s needed
so ..let it be

ben-kaufman:
I don't see the issue, the scheme is per DAO.

PR #338, opened 2019-09-24
Fix graph cli path

jellegerbrandy:
This is a small fix so that the deploy script will also be callable when called when teh subgraph is a dependency (as in `test_env`)


ops/mappings.json
```diff
@@ -1,7 +1,6 @@
 {
-  "private":
-    {
-      "mappings": [
+  "private": {
```
orenyodfat: why this file is changed ?

ops/graph-cli.js
```diff
@@ -1,41 +1,42 @@
-const spawn = require('spawn-command')
-const path = require('path')
+const spawn = require('spawn-command');
```
orenyodfat: looks like a lot of changes ?
is the lint ?

jellegerbrandy:
@orenyodfat fixes yr issues. Would be nice to approve so I can use this package :-)

PR #348, opened 2019-10-02
migration v28 and new rinkeby necdao

ben-kaufman:


daos/rinkeby/nectardao.json
```diff
@@ -0,0 +1,40 @@
...
+      "address": "0xC909620470eB665113F5cf401a461AeE433B39f2"
+    },
+    {
+      "name": "ReputationFromToken",
+      "alias": "ReputationFromToken",
+      "address": "0xCe934faaf5ba93AB246629e3Bc409d82Ad4fce78"
+    }
+  ],
+  "arcVersion": "0.0.1-rc.24"
+}
```
orenyodfat: why necdao is from arc 24 ? 
can we do another one from latest ?28 ?

daos/rinkeby/nectardao.json
```diff
@@ -0,0 +1,40 @@
...
+      "alias": "GENAuction",
+      "address": "0x1f75634317f108dAeF3Db3B6bedc767667464583"
+    },
+    {
+      "name": "ReputationFromToken",
+      "alias": "ReputationFromToken",
+      "address": "0x8401576723F4c4840f465b30e1F0a4A51508874c"
+    }
+  ],
+  "arcVersion": "0.0.1-rc.28"
```
orenyodfat: do we have universal scheme mapping for 28 ?

ben-kaufman: No but the addresses of the Universal schemes are the same as v24 so it shouldn't matter.

docker-compose.yml
```diff
@@ -31,6 +31,6 @@ services:
     environment:
       POSTGRES_PASSWORD: letmein
   ganache:
-    image: 'daostack/migration:0.0.1-rc.27-v0'
+    image: 'daostack/migration:0.0.1-rc.28-v0'
```
orenyodfat: should we use migration 0.0.1-rc.28-v1 ?

PR #350, opened 2019-10-03
Use latest arc on private network.

ben-kaufman:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@daostack/subgraph",
-  "version": "0.0.28",
+  "version": "0.0.29",
```
orenyodfat: why increment version ?

ben-kaufman: Because we said we want to release a new package.

PR #351, opened 2019-10-03
Added numberOfExpiredInQueueProposals

### LINKED ISSUE #344: add numberOfExpiredInQueueProposals
We'd like this both on "Scheme" and on "DAO" objects, similar to `numberOfBoostedProposals` etc.

ben-kaufman:
Fix #344

### DIFF
--- a/test/0.0.1-rc.16/domain.spec.ts
+++ b/test/0.0.1-rc.16/domain.spec.ts
@@ -1097,6 +1097,19 @@ describe('Domain Layer', () => {
       }
     }`;
 
+    const { proposalId: expiredInQueueProposal } = await propose({
+      rep: 10,
+      tokens: 10,
+      eth: 0,
+      external: 0,
+      periodLength: 0,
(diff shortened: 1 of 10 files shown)


test/0.0.1-rc.16/domain.spec.ts
```diff
@@ -551,6 +552,7 @@ describe('Domain Layer', () => {
         address: addresses.ContributionReward.toLowerCase(),
         name: 'ContributionReward',
         numberOfBoostedProposals: '0',
+        numberOfExpiredInQueueProposals: '0',
```
orenyodfat: is there a test which actually test that numberOfExpiredInQueueProposals increments sand decrements ?

ben-kaufman: I guess no. So I can add a proposal in the domain test.

orenyodfat: please add

PR #356, opened 2019-10-07
Added tags support

### LINKED ISSUE #354: new feature: tags
We are introducing tags in alchemy - basically a many-to-many relation between proposals and tag names.

We will implement this by:

- saving tags as a list on the ipfs "description", which will look like:
` ` `
{
title: "some tittle" ,
description: " ...." ,
tags: ["tag1" , " tag2" , "tag2" ]
}` 
- for the UI, we need to be able to: (a) list the tags with each proposal (b) get the complete list of tags and (c) search for proposals by tag

These data need to be indexed by the subgraph The most straightfoward solution seems to be:

- have a new `Tag` entity with attributes `id`  and `name` (or we can even use the ` name` as the id)
- add a property `tags`  to the Proposal entity, which refres to a list of Tag instances

ben-kaufman:
Fix #354

### DIFF
--- a/src/domain/proposal.ts
+++ b/src/domain/proposal.ts
@@ -66,12 +66,15 @@ export function getProposalIPFSData(proposal: Proposal): Proposal {
           proposal.url = descJson.toObject().get('url').toString();
         }
         let tags: string[] = [];
-        if (descJson.toObject().get('tags') != null && descJson.toObject().get('tags').kind === JSONValueKind.ARRAY) {
-          let tagsObjects = descJson.toObject().get('tags').toArray();
+        let tagsData =  descJson.toObject().get('tags');
+        if (tagsData != null && tagsData.kind === JSONValueKind.ARRAY) {
+          let tagsObjects = tagsData.toArray();
           for (let i = 0; i < tagsObjects.length; i++) {
             tags.push(tagsObjects[i].toString());
             let tagEnt = new Tag(tagsObjects[i].toString());
-            tagEnt.save();
+            if (Tag.load(tagsObjects[i].toString()) == null) {
+              tagEnt.save();
+            }
           }
           proposal.tags = tags;
         }
--- a/src/domain/proposal.ts
+++ b/src/domain/proposal.ts
@@ -1,6 +1,6 @@
 import { Address, BigDecimal, BigInt, Bytes, crypto, ipfs, json, JSONValueKind, store } from '@graphprotocol/graph-ts';
 import { GenesisProtocol } from '../types/GenesisProtocol/GenesisProtocol';
-import { ControllerScheme, DAO, GenesisProtocolParam, Proposal } from '../types/schema';
+import { ControllerScheme, DAO, GenesisProtocolParam, Proposal, Tag } from '../types/schema';
 import { concat, equalsBytes, equalStrings } from '../utils';
 import { getDAO, saveDAO } from './dao';
 import { updateThreshold } from './gpqueue';
@@ -64,6 +64,16 @@ export function getProposalIPFSData(proposal: Proposal): Proposal {
         }
         if (descJson.toObject().get('url') != null) {
           proposal.url = descJson.toObject().get('url').toString();
+        }
+        let tags: string[] = []
+        if (descJson.toObject().get('tags') != null && descJson.toObject().get('tags').kind === JSONValueKind.ARRAY) {
+          let tagsObjects = descJson.toObject().get('tags').toArray();
+          for (let i = 0; i < tagsObjects.length; i++) {
+            tags.push(tagsObjects[i].toString());
+            let tagEnt = new Tag(tagsObjects[i].toString());
+            tagEnt.save();
+          }
+          proposal.tags = tags;
         }
       }
     }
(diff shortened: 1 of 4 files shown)


jellegerbrandy:
This looks good but is only half of the spec from issue #354. 

`have a new Tag entity with attributes id and name (or we can even use the name as the id)` 

We need a way to get a list of all the tags, probably by

jellegerbrandy:
Sorry about that, I somehow overlooked half of the code (?), for me this is good to go (although it would be even better with a test that queryes the `tags`  collection :-)

src/domain/proposal.ts
```diff
@@ -1,6 +1,6 @@
 import { Address, BigDecimal, BigInt, Bytes, crypto, ipfs, json, JSONValueKind, store } from '@graphprotocol/graph-ts';
 import { GenesisProtocol } from '../types/GenesisProtocol/GenesisProtocol';
-import { ControllerScheme, DAO, GenesisProtocolParam, Proposal } from '../types/schema';
+import { ControllerScheme, DAO, GenesisProtocolParam, Proposal, Tag } from '../types/schema';
```
orenyodfat: where `Tag` is being used ?

src/domain/proposal.ts
```diff
@@ -65,6 +65,16 @@ export function getProposalIPFSData(proposal: Proposal): Proposal {
         if (descJson.toObject().get('url') != null) {
           proposal.url = descJson.toObject().get('url').toString();
         }
+        let tags: string[] = [];
+        if (descJson.toObject().get('tags') != null && descJson.toObject().get('tags').kind === JSONValueKind.ARRAY) {
+          let tagsObjects = descJson.toObject().get('tags').toArray();
+          for (let i = 0; i < tagsObjects.length; i++) {
```
orenyodfat: does the length should be capped somehow by max number of tags per proposal ? other wise this code might be fragile .

src/domain/proposal.ts
```diff
@@ -65,6 +65,16 @@ export function getProposalIPFSData(proposal: Proposal): Proposal {
...
+            tags.push(tagsObjects[i].toString());
+            let tagEnt = new Tag(tagsObjects[i].toString());
```
orenyodfat: how about check if already exist such a Tag  ? to avoid redundant write to db.

src/domain/schema.graphql
```diff
@@ -74,6 +74,7 @@ type Proposal @entity {
 	title: String
 	description: String
 	url: String
+	tags: [String!]
```
orenyodfat: tags: [Tag!] ?

src/domain/proposal.ts
```diff
@@ -65,6 +65,16 @@ export function getProposalIPFSData(proposal: Proposal): Proposal {
         if (descJson.toObject().get('url') != null) {
           proposal.url = descJson.toObject().get('url').toString();
         }
+        let tags: string[] = [];
+        if (descJson.toObject().get('tags') != null && descJson.toObject().get('tags').kind === JSONValueKind.ARRAY) {
+          let tagsObjects = descJson.toObject().get('tags').toArray();
```
orenyodfat: looks a bit cleaner  : 
```
let tagsData =  descJson.toObject().get('tags');
if (tagsData != null &&
      tagsData.kind === JSONValueKind.ARRAY) {
let tagsObjects = tagsData.toArray(); 
```

src/domain/proposal.ts
```diff
@@ -65,6 +65,19 @@ export function getProposalIPFSData(proposal: Proposal): Proposal {
...
           proposal.url = descJson.toObject().get('url').toString();
         }
+        let tags: string[] = [];
+        let tagsData =  descJson.toObject().get('tags');
+        if (tagsData != null && tagsData.kind === JSONValueKind.ARRAY) {
+          let tagsObjects = tagsData.toArray();
+          for (let i = 0; i < tagsObjects.length; i++) {
+            tags.push(tagsObjects[i].toString());
+            let tagEnt = new Tag(tagsObjects[i].toString());
+            if (Tag.load(tagsObjects[i].toString()) == null) {
```
orenyodfat: this check should move to line 73 ?

ben-kaufman: No... we want the tag in the list of the proposal anyway

src/domain/proposal.ts
orenyodfat: @jellegerbrandy does capping that with 100 tags per proposals sound reasonable ?

orenyodfat: @ben-kaufman please cap the number of tags per proposal with the magical number 100

src/domain/proposal.ts
```diff
@@ -70,6 +70,9 @@ export function getProposalIPFSData(proposal: Proposal): Proposal {
         if (tagsData != null && tagsData.kind === JSONValueKind.ARRAY) {
           let tagsObjects = tagsData.toArray();
           for (let i = 0; i < tagsObjects.length; i++) {
+            if (i === 100) {
+              break;
+            }
```
orenyodfat: `let tagsObjects = tagsData.toArray();
  let tagsLength = tagsObjects.length < 100 ?  tagsObjects.length : 100; 
 for (let i = 0; i < tagsLength; i++) {
...
`

orenyodfat: cleaner ?

PR #362, opened 2019-10-18
Add festDAO and requesting re-index

orishim:


daos/mainnet/fest-dao.json
```diff
@@ -0,0 +1,8 @@
+{
+  "arcVersion": "0.0.1-rc.23",
```
orenyodfat: how about using arc "0.0.1-rc.30" ?. by updating to latest migration

dOrgJelli: Hey @orenyodfat, this DAO was deployed using the DAOcreator, which is currently using v23 of @daostack/migration. I'll be sure to upgrade the dependency in the near future, and run some tests to make sure everything works. Some thoughts:
* Can we merge this, or does it need to be a "latest" arc version?
* Right now it's hard to know when new versions of Arc are released, how can we solve this? Should I fix my notifications on my side, or should DAOstack have some mailing list of sorts?
* Supporting new versions of Arc in the DAOcreator is not "automatic", as the migration script + resources are baked into the application's bundle. It's an interesting thought experiment to think about how we could make this "use latest version" process automatic... I don't think the deployment JS script should be a dynamic resource the app fetches, but the ABIs could be. If we standardize dao-params.json & the deployment script, and then make everything else data driven, we could achieve this IMO.

orenyodfat: > Can we merge this, or does it need to be a "latest" arc version? 
it can be merged though it is preferable (not a must) to use latest updated arc .

> Right now it's hard to know when new versions of Arc are released, how can we solve this? Should I fix my notifications on my side, or should DAOstack have some mailing list of sorts?

I guess that you can register to notification for new `migration` package release at npm.

> Supporting new versions of Arc in the DAOcreator is not "automatic", as the migration script + resources are baked into the application's bundle. It's an interesting thought experiment to think about how we could make this "use latest version" process automatic... I don't think the deployment JS script should be a dynamic resource the app fetches, but the ABIs could be. If we standardize dao-params.json & the deployment script, and then make everything else data driven, we could achieve this IMO.


one of the option can be that the application (daocreator) will import latest `migration` package (it will need to be inform of that in some way) . then it can do some devops to bake the dao-params.json  and use the already exist migration script (via the package). . for reference see work done at https://github.com/daostack/nectardao .
we can also have kind of `deployment` service  which  will always use the latest `migration` and expose an end point to send the dao-params.json

dOrgJelli: > it can be merged though it is preferable (not a must) to use latest updated arc .

Okay, I think we want to merge this.

> I guess that you can register to notification for new migration package release at npm.

Okay this was definitely on my end, turns out I had email notifications turned off for GitHub releases, sorry about that.

> ... automatic upgrade conversation ...

Yes this all sounds very interesting and doable. The one thing that is still a worry is that the dao-params.json's schema could be updated at any time, and DAOcreator wouldn't know this. I think there are a few things we can do to help solve this:
* Have the migration repo define TypeScript type definitions for the schema, this way compiler errors will stop DAOcreator from generating an incorrect dao-params.json.
* Have a versioning + upgrade system for the dao-params.json schema. This way, when the migration script is given an old version of a dao-params.json file ( v0.5 for example), it can run it through incremental upgrade scripts that turn it into the latest version (v0.9 for example), and then it knows for sure that the schema is what it's expecting. Major updates can still be made that break backwards compatibility, but it'd be up to the upgrade script to determine that and report back the error state.

orenyodfat: lets move this discussion to an issue at the migration repo .

PR #363, opened 2019-10-20
Fix Graphnode bugs

### LINKED ISSUE #361: Mappings use a `FirstRegisterScheme` type that is not in the GraphQL schema
The `Controller` and `UController` mappings both use a `FirstRegisterScheme` type that is not part of the subgraph's GraphQL schema. This worked so far, since we never validated the entity types that were loaded/saved with `store.get`/`store.set`. The upcoming version of graph-node will add validation for these and will cause the subgraph to fail.

The easiest fix would be to add the `FirstRegisterScheme` type to the schema and use the generated class instead of `store.get`/`store.set` directly.

Places where this type is used:

https://github.com/daostack/subgraph/blob/1bfe764903aeeff0908e0e70063f44cf638c8860/src/mappings/Controller/mapping.ts#L181-L193

https://github.com/daostack/subgraph/blob/1bfe764903aeeff0908e0e70063f44cf638c8860/src/mappings/UController/mapping.ts#L176-L188

ben-kaufman:
fix #361 
and 
validate that data sources need to have different names

### DIFF
--- a/ops/generate-daos-subgraph.js
+++ b/ops/generate-daos-subgraph.js
@@ -4,14 +4,26 @@ const { migrationFileLocation: defaultMigrationFileLocation, network } = require
 const {   subgraphLocation: defaultSubgraphLocation } = require('./graph-cli')
 const path = require("path");
 const currentDir = path.resolve(`${__dirname}`)
+let ids = [];
 
 function daoYaml(contract, contractAddress, arcVersion) {
   const { abis, entities, eventHandlers } = yaml.safeLoad(
     fs.readFileSync(`${currentDir}/../src/mappings/${contract}/datasource.yaml`, "utf-8")
   );
+
+  let name = contract;
+  if (ids.indexOf(contract) == -1) {
+    ids.push(contract);
+  } else if (ids.indexOf(contract+contractAddress) == -1) {
+    ids.push(contract+contractAddress);
+    name = contract+contractAddress;
+  } else {
+    return;
+  }
+
   return {
     kind: "ethereum/contract",
-    name: `${contract}`,
+    name: `${name}`,
     network: `${network}`,
     source: {
       address: contractAddress,
@@ -45,6 +57,14 @@ function daoYaml(contract, contractAddress, arcVersion) {
 async function generateSubgraph(opts={}) {
   opts.migrationFile = opts.migrationFile || defaultMigrationFileLocation;
   opts.subgraphLocation = opts.subgraphLocation || defaultSubgraphLocation;
+  const subgraphYaml = yaml.safeLoad(
+    fs.readFileSync(opts.subgraphLocation, "utf8")
+  );
+
+  for (var i = 0, len = subgraphYaml.dataSources.length; i < len; i++) {
+    ids.push(subgraphYaml.dataSources[i].name)
+  }
+
   let daodir
   if (opts.daodir) {
     daodir = path.resolve(`${opts.daodir}/${network}/`)
--- a/ops/generate-subgraph.js
+++ b/ops/generate-subgraph.js
@@ -52,6 +52,7 @@ async function generateSubgraph(opts={}) {
 }
 
 function combineFragments(fragments, isTemplate, addresses, missingAddresses) {
+  let ids = [];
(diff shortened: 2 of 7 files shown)


ops/generate-daos-subgraph.js
```diff
@@ -4,14 +4,26 @@ const { migrationFileLocation: defaultMigrationFileLocation, network } = require
...
+
+  let name = contract;
+  if (ids.indexOf(contract) == -1) {
+    ids.push(contract);
+  } else if (ids.indexOf(contract+contractAddress) == -1) {
+    ids.push(contract+contractAddress);
+    name = contract+contractAddress;
+  } else {
+    return;
+  }
```
orenyodfat: how about only  do `name = contract+ contractAddress` ?

ops/generate-subgraph.js
```diff
@@ -119,9 +120,19 @@ function combineFragments(fragments, isTemplate, addresses, missingAddresses) {
...
+    let name = contract;
+    if (ids.indexOf(contract) == -1) {
+      ids.push(contract);
+    } else if (ids.indexOf(contract+contractAddress) == -1) {
+      ids.push(contract+contractAddress);
+      name = contract+contractAddress;
+    } else {
+      return;
+    }
+
```
orenyodfat: how about only do `name = contract+ contractAddress` ?

src/mappings/Controller/mapping.ts
```diff
@@ -179,17 +180,12 @@ export function handleRegisterScheme(event: RegisterScheme): void {
...
+  let isFirstRegister = FirstRegisterScheme.load(avatar.toHex());
   if (isFirstRegister == null) {
     insertOrganization(event.address, avatar);
-    store.set(
-      'FirstRegisterScheme',
-      avatar.toHex(),
-      new Entity(),
-    );
+    isFirstRegister = new FirstRegisterScheme(avatar.toHex());
+    isFirstRegister.id = avatar.toHex();
```
orenyodfat: this line is redundant as the id is set be default at line 186.

src/mappings/UController/mapping.ts
```diff
@@ -174,17 +175,12 @@ export function handleRegisterScheme(event: RegisterScheme): void {
...
+  let isFirstRegister = FirstRegisterScheme.load(event.params._avatar.toHex());
   if (isFirstRegister == null) {
     insertOrganization(event.address, event.params._avatar);
-    store.set(
-      'FirstRegisterScheme',
-      event.params._avatar.toHex(),
-      new Entity(),
-    );
+    isFirstRegister = new FirstRegisterScheme(event.params._avatar.toHex());
+    isFirstRegister.id = event.params._avatar.toHex();
```
orenyodfat: this line is redundant as the id is set be default at line 181.

orenyodfat:
are we sure this is the issue? the graph node version is still 0.15.1 on master.
what cause the graph node issue to appears ? @ben-kaufman 
https://github.com/daostack/subgraph/blob/master/docker-compose.yml#L4

ops/generate-daos-subgraph.js
ben-kaufman: We can't because the graph node `codegen` function generates the `types` code based on the name, and we use this from our own code, so we must have at least one entity which has the original name.

ops/generate-subgraph.js
ben-kaufman: Same as above...

src/mappings/Controller/mapping.ts
ben-kaufman: removed

src/mappings/UController/mapping.ts
ben-kaufman: removed

orenyodfat:
the graph node issue "validate that data sources need to have different names" introduce at 
https://github.com/graphprotocol/graph-cli/releases/tag/v0.15.1 

please update graph-cli to use 0.15.3

orenyodfat:
please use fixed version of graph-cli and graph-ts (remove ^)

ben-kaufman:
I fixed the versions

Issue #366, opened 2019-10-24
null value resolved errors in subgraph v30


jellegerbrandy:
So this is a bit hard to reproduce, sorry about that.

After upgrading  to subgraph v30, I am seeing new errors after creating a genericScheme proposals:

```
  GraphQL error: Null value resolved for non-null field `expiresInQueueAt`
    GraphQL error: Null value resolved for non-null field `genesisProtocolParams`
    GraphQL error: Null value resolved for non-null field `gpQueue`
    GraphQL error: Null value resolved for non-null field `proposer`
    GraphQL error: Null value resolved for non-null field `votingMachine`
```

Here is to error on travis: https://travis-ci.com/daostack/client/builds/133351245#L2313 (ignore the lint errors)

To reproduce this locally:
```
git clonse git@github.com:daostack/client.git
git checkout tags
docker-compose up -d
npm run test -- test/proposal-genericscheme.spec.ts
```

PR #368, opened 2019-10-24
add sanity test

orenyodfat:
to prevent issues like #366

### DIFF
--- a/test/sanity.spec.ts
+++ b/test/sanity.spec.ts
@@ -1,0 +1,15 @@
+
+describe('sanity', () => {
+
+  it('Sanity', async () => {
+   //validate mapping include latest arc on private network.
+    const mappings = require(`../ops/mappings.json`);
+    let versionExist = false;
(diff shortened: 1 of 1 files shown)


test/sanity.spec.ts
```diff
@@ -0,0 +1,15 @@
+
+describe('sanity', () => {
+
+  it('Sanity', async () => {
+   // validate ops/mappings.json include latest arc on private network.
+    const mappings = require(`../ops/mappings.json`);
+    let latestArcVersionExist = false;
+    for (let i = 0; i < mappings.private.mappings.length; i++) {
+        if (mappings.private.mappings[i].arcVersion === '0.0.1-rc.30') {
```
ben-kaufman: We can just use the version in the `package.json` instead of hard-coding it. Please use this: `require('./package.json').dependencies['@daostack/arc']`.

orenyodfat: already done that

Issue #371, opened 2019-10-25
numberOfBoostedProposals is -1


jellegerbrandy:
numberOfBoostedProposals is -1, which is obviously wrong.

Run this query: 

```
{
  daos (where: {id: "0x0b93ba560283350d4216f29dc57e15df38d0eace"}) {
    id
    schemes {
      id
      numberOfBoostedProposals
    }
  }
}
```
on mainnet graph (for example here: https://thegraph.com/explorer/subgraph/daostack/v29_0)

orenyodfat:
not fixed yet.
#374 suppose to fix it

jellegerbrandy:
hey oren, this cannot be fixed in alchemy until it is deployed. So I'm re-opening this, if you don't mind :-)

orenyodfat:
the issue is fixed .
deploying the subgraph and alchemy readiness is another issue.

PR #373, opened 2019-10-27
v30-2

ben-kaufman:


orenyodfat:
Let’s wait to include the fix for #371

PR #379, opened 2019-10-30
Add Events for personal feed

### LINKED ISSUE #377: Personal feed requirements
To implement the personal feed in Alchemy we are going to need a new entity in the subgraph that contains all events that could show up in people's feeds. The events we need to track are listed here: https://docs.google.com/document/d/1_Qqoo5m9MbbuGyRG0HUchIFYPrDKU_LKr3T8mTambE4/edit#heading=h.n5awvhl2olm2

The entity can be called `Event`. We wil need to be able to filter events by the objects that they apply to (i.e. if a user is followign dao `0x124`, we want to be able to query all `Events` regarding that dao)

- `type`: one of list below
- `data!` (probably a JSON object with specific data for different types, like for a stake how much was staked - see list below)
- `proposal?` - id of the proposal that the event applies to
- `user?` - the address of the user that initiated the event (staker, voter, proposal creator..). 
- `dao?` - id of the DAO that the event applies to
- `time!` - when the event happened

Here is a list of event types that we want to see [with between square brackets the "index fields"  that are expected to be populated with this eventype:

- `NewProposal` [dao, user (address of proposer), proposal] - data contains `title` 
- `NewReputationHolder` [dao, user ] - data contains `reputationAmount`
- `ProposalStageChange`  [dao, proposal] - data contains the new `stage`
- `Vote` [proposal, user (=address of voter)] - data contains the `Outcome`  and `reputationAmount` (amount of rep)
- `Stake` [proposal, user (=address of staker)] - data contains `stakeAmount` and `Outcome`
-  `VoteFlip` [proposal, user (=address of voter)] - data contains the new `Outcome`, `votesFor`, `votesAgainst`
- `NewDAO` [dao] - data contina the dao's  `address` and  `name`

ben-kaufman:
Fix #377.

### DIFF
--- a/src/domain/event.ts
+++ b/src/domain/event.ts
@@ -112,7 +112,7 @@ function addEvent(
     proposal: string,
     user: Bytes,
     dao: string,
-    timestamp: BigInt
+    timestamp: BigInt,
 ): void {
     let eventEnt = new Event(id);
     eventEnt.type = type;
--- a/src/domain/event.ts
+++ b/src/domain/event.ts
@@ -1,0 +1,119 @@
+import { Address, BigInt, ByteArray, Bytes, crypto } from '@graphprotocol/graph-ts';
+import { Event, Proposal, ReputationHolder } from '../types/schema';
+import { concat } from '../utils';
+
+export function addNewDAOEvent(avatar: Address, daoName: string, timestamp: BigInt): void {
+    let eventType = 'NewDAO';
+
+    let eventEnt = new Event(avatar.toHex());
+    eventEnt.type = eventType;
+    eventEnt.data = '{ "address": "' + avatar.toHex() + '", "name": "' + daoName + '" }';
+    eventEnt.dao = avatar.toHex();
+    eventEnt.timestamp = timestamp;
+
+    eventEnt.save();
+}
+
+export function addProposalStateChangeEvent(proposalId: Bytes, timestamp: BigInt): void {
+    let proposal = Proposal.load(proposalId.toHex());
+    let eventType = 'ProposalStageChange';
+    let eventEntId = crypto.keccak256(concat(proposalId, timestamp as ByteArray));
+
+    let eventEnt = new Event(eventEntId.toHex());
+    eventEnt.type = eventType;
+    eventEnt.data = '{ "stage": "' + proposal.stage + '" }';
+    eventEnt.proposal = proposal.id;
+    eventEnt.dao = proposal.dao;
+    eventEnt.timestamp = timestamp;
+
+    eventEnt.save();
+}
+
+export function addNewReputationHolderEvent(reputationHolder: ReputationHolder): void {
+    let eventType = 'NewReputationHolder';
+    let eventEntId = crypto.keccak256(concat(reputationHolder.address, reputationHolder.createdAt as ByteArray));
+
+    let event = new Event(eventEntId.toHex());
+    event.type = eventType;
+    event.data = '{ "reputationAmount": "' + reputationHolder.balance.toString() + '" }';
+    event.user = reputationHolder.address;
+    event.dao = reputationHolder.dao;
+    event.timestamp = reputationHolder.createdAt;
+
+    event.save();
+}
+
+export function addVoteFlipEvent(proposalId: Bytes, proposal: Proposal, timestamp: BigInt): void {
+    let eventType = 'VoteFlip';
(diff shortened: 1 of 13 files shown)


src/domain/schema.graphql
```diff
@@ -188,3 +198,13 @@ type ContractInfo @entity {
 	version: String!
 	address: Bytes!
 }
+
+type Event @entity {
```
orenyodfat: how about adding events : [Event!] @derivedFrom('proposal') to proposal entity ?

src/domain/index.ts
```diff
@@ -273,6 +283,19 @@ export function handleStateChange(event: StateChange): void {
           (event.params._proposalState === 2)) {
           insertGPRewards(event.params._proposalId, event.block.timestamp, event.address, event.params._proposalState);
       }
+      let proposal = Proposal.load(event.params._proposalId.toHex());
+      let eventType = 'ProposalStageChange';
+      let eventEntId = crypto.keccak256(concat(event.params._proposalId, event.block.timestamp as ByteArray));
+
+      let eventEnt = new Event(eventEntId.toHex());
+      eventEnt.type = eventType;
+      eventEnt.data = '{ "stage": " ' + proposal.stage + ' " }';
```
orenyodfat: move that to event.ts

src/domain/schema.graphql
ben-kaufman: @jellegerbrandy will it be useful to add that?

src/domain/event.ts
```diff
@@ -0,0 +1,119 @@
+import { Address, BigInt, ByteArray, Bytes, crypto } from '@graphprotocol/graph-ts';
+import { Event, Proposal, ReputationHolder } from '../types/schema';
+import { concat } from '../utils';
+
+export function addNewDAOEvent(avatar: Address, daoName: string, timestamp: BigInt): void {
```
orenyodfat: how about add a function addEvent(address, name,id...) .
- so no need specific function and duplicate code for each event

ben-kaufman: done

src/domain/schema.graphql
```diff
@@ -188,3 +198,13 @@ type ContractInfo @entity {
...
+	id: ID!
+	type: EventType!
+	data: String!
```
orenyodfat: what is this data field ? why it is needed ?

ben-kaufman: That include some JSON data Alchemy might need. @jellegerbrandy

orenyodfat: @jellegerbrandy ?

src/domain/event.ts
```diff
@@ -0,0 +1,125 @@
...
+    addEvent(
+        'NewDAO',
+        avatar.toHex(),
+        '{ "address": "' + avatar.toHex() + '", "name": "' + daoName + '" }',
```
orenyodfat: what is that ?
why it is hard coded here?

ben-kaufman: That’s just the data object Alchemy needs for this event type...

orenyodfat: on first glance it looks not so elegant solution. is that can be avoided ? @jellegerbrandy

orenyodfat:
Well, is that cannot be analyzed from the event it self ?

src/domain/schema.graphql
orenyodfat: is that cannot be analyzed from the event it self ?

orenyodfat: @jellegerbrandy

src/domain/event.ts
jellegerbrandy: tibet is right: the avatar is not strictly needed (although neither does it hurt to have it :-))

I do not really understand @orenyodfat's objection. 
Events have a `data` attribute that is a string. This string happends to be a JSON structure, which is useful when we want to reprseent the event. This does not seem particularly unelegant to me as a pattern

orenyodfat:
Does this data need to be searchable and indexed?
It looks too specific and duplicate the field we  have on the entity it
self .

src/domain/event.ts
orenyodfat: Does this data need to be searchable and indexed?
It looks too specific and duplicate the field we  have on the entity it self .

jellegerbrandy: no it does not, that is why it is in a text field and not in its own separate field.

PR #380, opened 2019-11-01
DAOTracker Support

dOrgJelli:


ops/mappings.json
```diff
@@ -610,6 +610,13 @@
         "dao": "base",
         "mapping": "Redeemer",
         "arcVersion": "0.0.1-rc.31"
+      },
+      {
+        "name": "DAOTracker",
+        "contractName": "DAOTracker",
+        "dao": "base",
+        "mapping": "DAOTracker",
+        "arcVersion": "0.0.1-rc.29"
```
orenyodfat: it should be 0.0.1-rc.31

src/mappings/DAOTracker/datasource.yaml
```diff
@@ -0,0 +1,20 @@
...
+  - DAOTrackerContract
+templates:
+  - mapping: Avatar
+    version: 0.0.1-rc.31
+  - mapping: Controller
+    version: 0.0.1-rc.31
+  - mapping: DAOToken
+    version: 0.0.1-rc.31
+  - mapping: Reputation
+    version: 0.0.1-rc.31
```
orenyodfat: where the schemes templates (e.g GenericScheme) is added ?

src/mappings/DAOTracker/mapping.ts
```diff
@@ -0,0 +1,127 @@
...
+  ResetDAO,
+  TrackDAO,
+} from '../../types/DAOTracker/DAOTracker';
+import {
+  BlacklistedDAO,
+  DAOTrackerContract,
+  ResetDAO as ResetDAOEntity,
+} from '../../types/schema';
+import {
+  Avatar_0_0_1_rc_31,
```
orenyodfat: what is this postfix is for ? Avatar_0_0_1_rc_31 ?

ops/generate-subgraph.js
```diff
@@ -130,6 +132,12 @@ function combineFragments(fragments, isTemplate, addresses, missingAddresses) {
       return;
     }
 
+    if (isTemplate) {
+      // convert name to be version specific
+      const classVersion = version.replace(/\.|-/g, '_');
```
orenyodfat: how about generate a pre compiled mapping from for all specific version template ?
this will make the code more readable and maintainable.

src/mappings/DAOTracker/mapping.ts
```diff
@@ -0,0 +1,127 @@
...
+  Controller_0_0_1_rc_31,
+  DAOToken_0_0_1_rc_31,
+  Reputation_0_0_1_rc_31,
+} from '../../types/templates';
+import { equalStrings } from '../../utils';
+
+export function getDAOTrackerContract(address: Address): DAOTrackerContract {
+  let daoTracker = DAOTrackerContract.load(address.toHex()) as DAOTrackerContract;
```
orenyodfat: does the DAOTrackerContract is used by any anyone ? is that really needed ?

src/mappings/DAOTracker/mapping.ts
```diff
@@ -0,0 +1,127 @@
...
+export function handleTrackDAO(event: TrackDAO): void {
+  // Ensure the DAOTrackerContract has been added to the store
+  getDAOTrackerContract(event.address);
+
+  let _avatar = event.params._avatar;
+  let _controller = event.params._controller;
+  let _reputation = event.params._reputation;
+  let _daoToken = event.params._daoToken;
+
+  /* TODO: uncomment when this issue is resolved https://github.com/graphprotocol/graph-node/issues/1333
```
orenyodfat: this issue is closed

src/mappings/DAOTracker/mapping.ts
```diff
@@ -0,0 +1,127 @@
...
+  // If the avatar hasn't been blacklisted
+  let daoTrackerSC = DAOTracker.bind(event.address);
+  if (daoTrackerSC.blacklisted(_avatar), 'latest') {
+    return;
+  }
+  */
+
+  // If the avatar already exists, early out
+  if (store.get('AvatarContract', _avatar.toHex()) != null) {
```
orenyodfat: if (AvatarContract.load(_avatar.toHex()) != null) {
..

src/mappings/DAOTracker/mapping.ts
```diff
@@ -0,0 +1,127 @@
...
+  Avatar_0_0_1_rc_31.create(_avatar);
+  Reputation_0_0_1_rc_31.create(_reputation);
+  DAOToken_0_0_1_rc_31.create(_daoToken);
+
+  // Track the Controller if it isn't a UController we're already tracking
+  if (store.get('UControllerOrganization', _controller.toHex()) == null) {
+    Controller_0_0_1_rc_31.create(_controller);
+  }
+
+  // Note, no additional work is needed here because...
```
orenyodfat: how about tracking the dao schemes ( e.g GenericScheme) ?

src/mappings/DAOTracker/mapping.ts
```diff
@@ -0,0 +1,127 @@
...
+  if (store.get('ResetDAO', _avatar.toHex())) {
+    store.remove('ResetDAO', _avatar.toHex());
+  }
+}
+
+export function handleResetDAO(event: ResetDAO): void {
+  // Ensure the DAOTrackerContract has been added to the store
+  getDAOTrackerContract(event.address);
+
+  let _avatar = event.params._avatar;
```
orenyodfat: why local functions vars starts with _ ?

src/mappings/DAOTracker/mapping.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  if (!equalStrings(_explanationHash, '')) {
+    let explanation = ipfs.cat('/ipfs/' + _explanationHash);
+
+    if (explanation != null) {
+      blacklistedDAO.explanation = explanation.toString();
+    }
+  }
+
+  store.set('BlacklistedDAO', blacklistedDAO.id, blacklistedDAO);
```
orenyodfat: BlacklistedDAO.save()

src/mappings/DAOTracker/mapping.ts
```diff
@@ -0,0 +1,127 @@
...
+
+export function handleBlacklistDAO(event: BlacklistDAO): void {
+  // Ensure the DAOTrackerContract has been added to the store
+  let daoTracker = getDAOTrackerContract(event.address);
+
+  let _avatar = event.params._avatar;
+  let _explanationHash = event.params._explanationHash;
+
+  // Add the BlacklistedDAO to the store
+  let blacklistedDAO = new BlacklistedDAO(_avatar.toHex());
```
orenyodfat: where does the blacklist is check ? and how it is implemented ?
I guess that on first phase , we should have a pre compiled black list daos .
subgraph will need to be reindexed to ignore the blacklisted dao

src/mappings/DAOTracker/mapping.ts
orenyodfat: I guess we need to have a pre compiled Daos black list.
the subgraph will need to be re indexed to exclude the blacklist dao .(at least on the first phase of the implementation ).

src/mappings/Reputation/datasource.yaml
```diff
@@ -10,3 +10,5 @@ eventHandlers:
     handler: handleMint
   - event: Burn(indexed address,uint256)
     handler: handleBurn
+  - event: OwnershipTransferred(indexed address,indexed address)
```
orenyodfat: why is that needed ?

ops/mappings.json
dOrgJelli: The contract was introduced in rc.29, and the address & ABI haven't changed since, so I think it should be rc.29 no?

src/mappings/DAOTracker/datasource.yaml
dOrgJelli: The scheme contracts do not need to be templates because their addresses are known at startup. We cannot support dynamically tracking schemes because we don't currently have a way of determining the "Scheme Type" when a new scheme is added to the controller. We will once Arc Factories are a thing though.

src/mappings/DAOTracker/mapping.ts
dOrgJelli: These names are currently auto generated [here](https://github.com/dOrgTech/subgraph/blob/ab7135abeb86ad27997b1d92acb7811ba8fc6ef9/ops/generate-subgraph.js#L137) based on the version stated [here](https://github.com/dOrgTech/subgraph/blob/ab7135abeb86ad27997b1d92acb7811ba8fc6ef9/src/mappings/DAOTracker/datasource.yaml#L7).

The reason is because datasource templates cannot have the same name, and we need a way of supporting multiple versions of the ABI.

ops/generate-subgraph.js
dOrgJelli: I'm not sure what you mean by this.

src/mappings/DAOTracker/mapping.ts
dOrgJelli: I think it's necessary for administrative debugging / checking. Also in the UI if a user tries to view a DAO that's been blacklisted, we'd want to be able to show the reason as to why it's been blacklisted from the UI.

src/mappings/DAOTracker/mapping.ts
dOrgJelli: Yes we'll need to have the precompiled list instead of checking this at "index time" since we can't view the "latest" version of the registry :/

dOrgJelli: I'll remove this from the code. This also means that we don't need an on-chain mapping of address=>boolean, and we can instead just use the emitted events.

dOrgJelli: Should we remove this from the contract and re-deploy?

src/mappings/DAOTracker/mapping.ts
dOrgJelli: Fixed

src/mappings/DAOTracker/mapping.ts
dOrgJelli: We have no way of determining what the type of the scheme is when it's added to the controller.

src/mappings/DAOTracker/mapping.ts
dOrgJelli: Whoops, apologies, fixing

src/mappings/Reputation/datasource.yaml
dOrgJelli: I thought it made sense since DAOToken and Avatar track this event and instantate the entity in the store based off of it. Also I htink there could be cases where a new DAO is deployed without any initial REP holders, and we would still want the ReputationContract entity to exist in the store for completeness sake.

A good example of this is when you have a token holder DAO that relies on token holders redeeming reputation. When the DAO is first deployed there may be zero REP holders.

src/mappings/DAOTracker/mapping.ts
dOrgJelli: See comment above. Yes we'll need to have a precompiled list and reindex to reflect that, and Alchemy could query the blacklist to see if it wants to show a DAO and also be able to show the reason why the DAO was blacklisted.

ops/mappings.json
orenyodfat: the private network (ganache) mapping should include all latest version contracts.
(need to set the number to 18 here https://github.com/daostack/subgraph/blob/master/test/sanity.spec.ts#L15 )

src/mappings/DAOTracker/datasource.yaml
orenyodfat: well, you can determining the `Scheme Type` via contractInfo , via its address

src/mappings/DAOTracker/mapping.ts
orenyodfat: is that can be avoided somehow by getting the template from a pre defined mapping or something ?

src/mappings/DAOTracker/mapping.ts
orenyodfat: can be read from the contractInfo

src/mappings/DAOTracker/mapping.ts
```diff
@@ -0,0 +1,121 @@
...
+  let daoToken = event.params._daoToken;
+
+  // If the avatar already exists, early out
+  if (AvatarContract.load(avatar.toHex()) != null) {
+    return;
+  }
+
+  // Tell the subgraph to start indexing events from the:
+  // Avatar, Controller, DAOToken, and Reputation contracts
+  Avatar_0_0_1_rc_31.create(avatar);
```
orenyodfat: how about : 
- create a template_create module which hold all templates history.
then the code should looks like:  
```
let contractInfo = ContractInfo.load(avatar.toHex());
  if (contractInfo != null) {
     tempate_create("avatar",contractInfo.version);
  }
```

src/mappings/DAOTracker/mapping.ts
orenyodfat: see https://github.com/daostack/subgraph/pull/380/files#r342098970 for a reference

ops/generate-subgraph.js
orenyodfat: see https://github.com/daostack/subgraph/pull/380/files#r342098970

src/mappings/DAOTracker/mapping.ts
dOrgJelli: Oooo I really like this approach. Will look into how to implement this.

src/mappings/DAOTracker/mapping.ts
dOrgJelli: Responded to your comment linked above, will look into this.

ops/generate-subgraph.js
dOrgJelli: Will look into this.

src/mappings/DAOTracker/mapping.ts
dOrgJelli: Yes but only for schemes we know the address of on startup.

Tracking non-universal schemes the subgraph already knows the address of at startup is handled.

dOrgJelli:
@orenyodfat tests are passing

src/mappings/DAOTracker/mapping.ts
dOrgJelli: Okay so this actually doesn't make sense @orenyodfat, since Avatar is a newly deployed Avatar that we don't know the address of at build-time. This means that it wouldn't be contained within ContractInfo.

We cannot know the ABI version of these newly added / tracked contracts for Avatar, Controller, DAOToken, and Reputation since we don't have factories that're spawning them to go off of.

src/mappings/DAOTracker/mapping.ts
dOrgJelli: Looked into it, and responded to your comment here: https://github.com/daostack/subgraph/pull/380/files#r342098970

ops/generate-subgraph.js
dOrgJelli: Looked into it and responded to your comment here, it's impossible to do this: https://github.com/daostack/subgraph/pull/380/files#r342098970

ops/mappings.json
dOrgJelli: Updated

src/mappings/DAOTracker/mapping.ts
orenyodfat: so we cannot be sure that the avatar is from v 0_0_1_rc_31 either. right ?
how about `TrackDAO` event will give us clue re arc version ?

- DAOCreator will be index and added to contractInfo.
- DAOTracker track will get arc version as a param : 
 `function track(Avatar _avatar, Controller _controller, string _arcVersion)`
- TrackDAO will emit the msg.sender and _arcVersion  along with all other addresses : 
`    event TrackDAO(address indexed _avatar, address _controller, address _reputation, address _daoToken, address _sender, string _arcVersion);
- if  _sender is DAOCreator (we can check that via contractInfo) so the avatar version is the DAOCreator version. - in this case the _arcVersion is ignored.
-  if  _sender is not DAOCreator the avatar version is _arcVersion.

dOrgJelli: https://github.com/daostack/arc/pull/683

dOrgJelli: https://github.com/daostack/migration/pull/219

dOrgJelli: Should the DAOTracker mapping support both the old & new TrackDAO events?

dOrgJelli:
@orenyodfat the new DAOTracker contract that contains the arcVersion is integrated. Here's an overview of the new changes:
* Added the `ops/templates.json` file to keep track of what ABI versions should be used with what mappings.
* Updated `generate-contractsInfos.js` to...
`setTemplateInfo('Avatar', '0.0.1-rc.16', 'Avatar_0_0_1_rc_16');`
for each template templatable ABI. `setTemplateInfo(...)` is defined in `src/utils.ts`.
* Updated `generate-subgraph.js` to add all possible datasource templates to the subgraph manifest.
* Updated `DAOTracker/mapping.ts` to use the new way of instantiating templates.

Open question:
* In `DAOTracker/mapping.ts`, at line # 70, how do you think we should handle the error case where a template isn't found? Should we just fail silently and return from the handler?

src/mappings/DAOTracker/mapping.ts
```diff
@@ -0,0 +1,146 @@
...
+  // Ensure the DAOTrackerContract has been added to the store
+  getDAOTrackerContract(event.address);
+
+  let avatar = event.params._avatar;
+  let controller = event.params._controller;
+  let reputation = event.params._reputation;
+  let daoToken = event.params._daoToken;
+  let sender = event.params._sender;
+  let arcVersion = event.params._arcVersion;
+
```
orenyodfat: these local vars are redundant here.

src/mappings/DAOTracker/mapping.ts
```diff
@@ -0,0 +1,146 @@
...
+  let avatarTemplate = fetchTemplateName('Avatar', arcVersion);
+  let controllerTemplate = fetchTemplateName('Controller', arcVersion);
+  let reputationTemplate = fetchTemplateName('Reputation', arcVersion);
+  let daoTokenTemplate = fetchTemplateName('DAOToken', arcVersion);
+
+  let missingTemplate = avatarTemplate == null ||
+                        reputationTemplate == null ||
+                        daoTokenTemplate == null;
+
+  let universalController = UControllerOrganization.load(controller.toHex()) != null;
```
orenyodfat: `!= null` ?

src/mappings/DAOTracker/mapping.ts
```diff
@@ -0,0 +1,146 @@
...
+  createTemplate(reputationTemplate, reputation);
+  createTemplate(daoTokenTemplate, daoToken);
+
+  // Track the Controller if it isn't a UController we're already tracking
+  if (universalController === false) {
+    createTemplate(controllerTemplate, controller);
+  }
+
+  // Note, no additional work is needed here because...
+  // * ControllerOrganization is added to the store by the 'RegisterScheme' event
```
orenyodfat: does the tests cover that ?
indexing a Dao with its schemes ?
how about none universal schemes ? like genesisProrocol ?

src/mappings/DAOTracker/mapping.ts
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  // If the sender of the 'track' call is the DaoCreator contract, use its arcVersion
+  let daoCreatorInfo = ContractInfo.load(sender.toHex());
+  if (daoCreatorInfo != null) {
+    if (equalStrings(daoCreatorInfo.name, 'DaoCreator')) {
+      arcVersion = daoCreatorInfo.version;
+    }
+  }
+
```
orenyodfat: need to confirm (somehow) that creating a template for wrong contract does not crash the subgraph .
this can happened only if the event does not come from daocreator.
if we cannot confirm ,so lets consider  limit this feature only for daocreator.

orenyodfat:
> @orenyodfat the new DAOTracker contract that contains the arcVersion is integrated. Here's an overview of the new changes:
> 
> * Added the `ops/templates.json` file to keep track of what ABI versions should be used with what mappings.
> * Updated `generate-contractsInfos.js` to...
>   `setTemplateInfo('ContractName', 'ArcVersion', 'UniqueTemplateName');`
>   for each templatable ABI. `setTemplateInfo(...)` is defined in `src/utils.ts`.
>   ex: `setTemplateInfo('Avatar', '0.0.1-rc.16', 'Avatar_0_0_1_rc_16');`
> * Updated `generate-subgraph.js` to add all possible datasource templates to the subgraph manifest.
> * Updated `DAOTracker/mapping.ts` to use the new way of instantiating templates.
> 
> Open question:
> 
> * In `DAOTracker/mapping.ts`, at line # 70, how do you think we should handle the error case where a template isn't found? Should we just fail silently and return from the handler? This essentially means (assuming correct on-chain data) the subgraph is out of date and is missing ABIs + mappings. I propose we add a new entity to the store when this happens, and our automation looks for this entity existing to alerts the team letting them know what's going wrong.

this can happened only when the event does not come from daocreator. right ? lets consider supporting this feature only if it come from daocreator.

src/mappings/DAOTracker/mapping.ts
dOrgJelli: If loading the UControllerOrganization != null, it means we're using the universalController. If we're not using the universal controller, we want to (1) check to see if we're missing the template for the controller version, and (2) create the template if it exists.

src/mappings/DAOTracker/mapping.ts
dOrgJelli: Yes the test currently adds a ContributionReward scheme and ensures it shows up. Non-universal schemes work as long as the subgraph knows about its address and type in the ContractInfos (at startup).

src/mappings/DAOTracker/mapping.ts
dOrgJelli: The subgraph does not (and cannot) check to make sure the contract bytecode matches the bytecode in the ABI file. This means that it only watches for events ABIs from a given contract address, and if they do not emit from that address the subgraph will not do anything.

So in the case you're describing, the subgraph would not crash when it creates the template, it would rather just do nothing from there on out. The only case where this could happen is someone wrote custom contracts that emit the events we're looking for, and the subgraph responds to those events from these newly created contracts. The plan here, is that if DAOstack detects a "DAO" is doing this, it would blacklist it.

dOrgJelli: Writing a test to verify the subgraph doesn't crash.

dOrgJelli:
@orenyodfat yes this only happens when it comes not from the DAOcreator. I agree that we could limit this to just the DAOcreator for now, and the work done here would still be reusable for when the factories are created.

src/mappings/DAOTracker/mapping.ts
orenyodfat: as far as I see the only way to mitigate that and stay on the safe side is to track this event only if it come via daocreator. otherwise ignore it.

src/mappings/DAOTracker/mapping.ts
orenyodfat: we need to find a way to index NonUni schemes as well without the subgraph knows its address and type in compile time.
we might need additional event for that from daotracker @dOrgJelli

src/mappings/DAOTracker/mapping.ts
orenyodfat: oh, the syntax is exceptional .
`let universalController = (UControllerOrganization.load(controller.toHex()) != null);` this might be more readable ..

src/mappings/DAOTracker/mapping.ts
```diff
@@ -0,0 +1,146 @@
...
+    daoTracker = new DAOTrackerContract(address.toHex());
+    daoTracker.address = address;
+    let daoTrackerSC = DAOTracker.bind(address);
+    daoTracker.owner = daoTrackerSC.owner();
+    daoTracker.save();
+  }
+  return daoTracker;
+}
+
+export function handleTrackDAO(event: TrackDAO): void {
```
orenyodfat: check for precompiled blacklist daos

dOrgJelli:
@orenyodfat The only DAOs that will be tracked are ones created by the DaoCreator. Additionally the embedded blacklist is implemented in the `ops/blacklist.json` file.

PR #383, opened 2019-11-03
Fix ttravis deploy

ben-kaufman:


ops/generate-subgraph.js
```diff
@@ -118,7 +118,7 @@ function combineFragments(fragments, isTemplate, addresses, missingAddresses) {
     } : {
       address: contractAddress,
       abi,
-      startBlock: startBlock
+      startBlock
```
orenyodfat: are you sure about that ?
how does it worked before ?

ben-kaufman: This is just a small tweak to make the code nicer, the actual fix is the `parseInt` in the settings file

PR #389, opened 2019-11-13
Increase travis deploy timeout

ben-kaufman:


orenyodfat:
does this pr is needed ?

PR #390, opened 2019-11-13
Bump graph-cli to 0.16.2 (fast IPFS upload)

Jannis:
@orenyodfat Little PR for you to speed up those `graph build` / `graph deploy` commands. Overall I think the whole `npm run deploy` procedure now takes approx. 100s instead of 30mins.

PR #391, opened 2019-11-14
Use latest migration

ben-kaufman:
use graph-cli 16.0.2 for fast upload

### DIFF
--- a/daos/private/testdao.json
+++ b/daos/private/testdao.json
@@ -1,5 +1,5 @@
 {
-  "name": "Sagacious Senor",
+  "name": "Obtuse Artisan",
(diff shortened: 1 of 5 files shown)


daos/private/testdao.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "Sagacious Senor",
+  "name": "Obtuse Artisan",
```
orenyodfat: does this file need to be updated or it can be added to .gitignore ?

orenyodfat:
use graph-cli v 0.16.2 as in #390 , @ben-kaufman

daos/private/testdao.json
ben-kaufman: Yes it needs as we use it in the tests.

PR #392, opened 2019-11-14
Minor Fix

dOrgJelli:


dOrgJelli:
@orenyodfat I believe this will fix the issue you're seeing on Rinkeby

src/mappings/DAOTracker/mapping.ts
```diff
@@ -22,6 +26,8 @@ export function getDAOTrackerContract(address: Address): DAOTrackerContract {
     daoTracker.owner = daoTrackerSC.owner();
     daoTracker.save();
     setBlacklistedDAOs();
+    setContractsInfo();
```
orenyodfat: why it is needed here ?

dOrgJelli: Because we're searching for the DaoCreator's ContractInfo here, and we must ensure it's in the store before checking: https://github.com/daostack/subgraph/blob/8ed8b36371b93c1dd02aee065a5b6fb89f51e35c/src/mappings/DAOTracker/mapping.ts#L57

orenyodfat: ok.this is needed here if there is no pre indexed daos . so this will be the first place.
👍

PR #394, opened 2019-11-19
Use migration v33

ben-kaufman:


orenyodfat:
please update 17 to 18 here https://github.com/daostack/subgraph/blob/master/test/sanity.spec.ts#L15

PR #400, opened 2019-12-01
Support Arc experimental

ben-kaufman:


docker-compose.yml
```diff
@@ -31,6 +31,6 @@ services:
     environment:
```
orenyodfat: - use graph node 0.16.1 https://github.com/graphprotocol/graph-node/releases/tag/v0.16.1
- IPFS max file size should be increased to 300000.

ops/generate-daos-subgraph.js
```diff
@@ -4,6 +4,7 @@ const { migrationFileLocation: defaultMigrationFileLocation, network, startBlock
 const {   subgraphLocation: defaultSubgraphLocation } = require('./graph-cli')
 const path = require("path");
 const currentDir = path.resolve(`${__dirname}`)
+const supportedSchemes = ['ContributionReward', 'GenericScheme', 'SchemeRegistrar']
```
orenyodfat: what is this ?

ops/mappings.json
```diff
@@ -2,623 +2,56 @@
...
-        "mapping": "DAOTracker",
-        "arcVersion": "0.0.1-rc.33"
-      }
-    ]
+        {
+          "name": "GEN",
+          "contractName": "GEN",
+          "dao": "package",
+          "mapping": "DAOToken",
+          "arcVersion": "0.0.1-rc.2"
```
orenyodfat: this arc version might be confusing with the legacy one .
maybe we should give it other name ? 
`1.0.1-rc.2` or `exp-0.0.1-rc.2`?

ops/utils.js
```diff
@@ -2,7 +2,7 @@ const fs = require("fs");
 
 function versionToNum(version) {
   const strlen = version.length;
-  return Number(version.slice(strlen - 2, strlen));
+  return Number(version.slice(strlen - 1, strlen));
```
orenyodfat: how about checking the Len of the version first and then decide if to do strlen-1 or strlen-2 ? this will works for both 2 and 1 digits rc versions.

src/domain/dao.ts
```diff
@@ -42,23 +42,11 @@ export function insertNewDAO(
   dao.nativeToken = nativeTokenAddress.toHex();
   dao.nativeReputation = nativeReputationAddress.toHex();
   dao.reputationHoldersCount = BigInt.fromI32(0);
-  dao.register = 'na';
   saveDAO(dao);
 
   return dao;
 }
 
-export function register(
```
orenyodfat: why is this removed?

src/domain/schema.graphql
```diff
@@ -29,7 +29,6 @@ type DAO @entity {
 	reputationHolders: [ReputationHolder!] @derivedFrom(field: "dao")
 	reputationHoldersCount: BigInt!
 	rewards: [GPReward!] @derivedFrom(field: "dao")
-	register: String!
```
orenyodfat: why?

src/mappings/DAOTracker/mapping.ts
```diff
@@ -12,8 +12,8 @@ import {
   AvatarContract,
```
orenyodfat: doa tracker should be removed.
we should track new daos  according to the daoFactory events.

orenyodfat:
increment package version to "exp-0.1.0-1",

ops/generate-daos-subgraph.js
ben-kaufman: To index properly the schemes which the subgraph support and implemented.

ops/mappings.json
ben-kaufman: Ok, shouldn't be a problem, but I'll need you to first rename the arc package like this.

src/domain/dao.ts
ben-kaufman: Beacuse we don't have DAORegistry with the new Arc.

src/domain/schema.graphql
ben-kaufman: Because we removed DAORegistry from Arc.

src/mappings/DAOTracker/mapping.ts
ben-kaufman: But DAOTracker is part of DAOFactory, DAOFactory uses it and accept it as an initialize parameter.

ops/generate-daos-subgraph.js
orenyodfat: why it is needed to set explicitly here. ? 
how about daoFactory ?

ben-kaufman: How is that related to the DAOFactory? This is related to which schemes the subgraph has mappings for and is indexing with full support, namely, the GenericScheme, ContributionReward, and SchemeRegistrar. It is not related to Arc or the DAOFactory in any way.

orenyodfat: why it is needed to set explicitly here. ?

src/domain/dao.ts
orenyodfat: DAORegistry was never part of arc . it comes from arc-hive repo https://github.com/daostack/arc-hive/blob/master/contracts/DAORegistry.sol .
we need that.

src/domain/schema.graphql
orenyodfat: DAORegistry was never part of arc . it comes from arc-hive repo https://github.com/daostack/arc-hive/blob/master/contracts/DAORegistry.sol .
we need that.

src/mappings/DAOTracker/mapping.ts
orenyodfat: will remove it  from daoFactory

ops/generate-daos-subgraph.js
ben-kaufman: So that we can know if we should generate entities for them or not.

src/domain/dao.ts
ben-kaufman: Oh ok. I thought it will be removed with the new Architecture.

src/mappings/DAOFactory/mapping.ts
```diff
@@ -0,0 +1,138 @@
...
+} from '../../contractsInfo';
+
+import { Avatar } from '../../types/Controller/Avatar';
+import { DAOFactory, InitialSchemesSet, NewOrg, ProxyCreated, SchemeInstance } from '../../types/DAOFactory/DAOFactory';
+import {
+  AvatarContract, ContractInfo, DAOFactoryContract,
+} from '../../types/schema';
+import { createTemplate, debug, fetchTemplateName, setContractInfo } from '../../utils';
+
+export function getDAOFactoryContract(address: Address): DAOFactoryContract {
```
orenyodfat: is this must be export function ?

src/mappings/DAOFactory/mapping.ts
```diff
@@ -0,0 +1,138 @@
...
+    daoFactory.app = daoFactoryContract.app();
+    daoFactory.save();
+    setContractsInfo();
+    setTemplatesInfo();
+  }
+  return daoFactory;
+}
+
+export function handleNewOrg(event: NewOrg): void {
+  // Ensure the DAOTrackerContract has been added to the store
```
orenyodfat: DAOFactory

src/mappings/DAOFactory/mapping.ts
```diff
@@ -0,0 +1,138 @@
...
+  getDAOFactoryContract(event.address);
+
+  let avatar = event.params._avatar;
+
+  // If the avatar already exists, early out
+  if (AvatarContract.load(avatar.toHex()) != null) {
+    return;
+  }
+
+  let avatarContract = Avatar.bind(avatar);
```
orenyodfat: does arc/package version is ignored here ?

src/mappings/DAOFactory/mapping.ts
```diff
@@ -0,0 +1,138 @@
...
+  let arcVersion: string;
+
+  let avatarInfo = ContractInfo.load(avatar.toHex());
+  if (avatarInfo != null) {
+    avatarInfo.name = 'Avatar';
+    avatarInfo.alias = 'Avatar';
+    avatarInfo.save();
+    arcVersion = avatarInfo.version;
+  } else {
+    // If contractInfo doesn't exist something went wrong and should exit early
```
orenyodfat: is that won't be the case on any new org created via dao factory ?

ben-kaufman: Not sure I understand the question, but the `ProxyCreated` event is fired before the NewOrg and thus make sure the contractInfo exists beforehand. This check is just to make sure the subgraph won't crash if somehow the contractInfo doesn't exist (though it always should).

src/mappings/DAOFactory/mapping.ts
ben-kaufman: Basically yes. We did not yet decide how we should deal with multi version support in the new subgraph so for now we still have this "version dependency".

src/mappings/DAOFactory/mapping.ts
orenyodfat: contractInfo is a preCompiled data. 
it does not hold avatar addresses.

orenyodfat: is there a test for this mapping ? and specific for this functionality ?

ben-kaufman: It can be precompiled, but not necessarily only so, and it also had `Avatar` and all other contracts. There's no reason not to use it here now that we only auto index daos.

There is a short test for that. https://github.com/daostack/subgraph/blob/subgraph-experimental/test/0.1.1-rc.1/contractsInfo.spec.ts

We could add more though doesn't seem necessary as this is a very simple mapping.

PR #402, opened 2019-12-03
Use migration v34

ben-kaufman:


ops/generate-daos-subgraph.js
```diff
@@ -38,6 +38,12 @@ function daoYaml(contract, contractAddress, arcVersion) {
       entities,
       abis: (abis || [contract]).map(contractName => {
         let _arcVersion = Number(arcVersion.slice(arcVersion.length-2,arcVersion.length));
+        if ((_arcVersion >= 34) && (contractName === "UGenericScheme")) {
```
orenyodfat: UGenericScheme is deprecated on 34 . no ?

ops/generate-subgraph.js
```diff
@@ -71,6 +71,12 @@ function combineFragments(fragments, isTemplate, addresses, missingAddresses) {
       abis = (yamlLoad.abis || [contract]).map(contractName => {
         const versionNum = versionToNum(version);
 
+        if ((versionNum >= 34) && (contractName === "UGenericScheme")) {
```
orenyodfat: UGenericScheme is deprecated on 34 . no ?

ben-kaufman: Yes, that's why if a DAO still uses it I have to bring it's ABI from v33

orenyodfat: we should return error in this case

ops/mappings.json
```diff
@@ -527,96 +555,117 @@
...
-        "arcVersion": "0.0.1-rc.33"
+        "arcVersion": "0.0.1-rc.34"
       },
       {
         "name": "DAOTracker",
         "contractName": "DAOTracker",
         "dao": "base",
         "mapping": "DAOTracker",
-        "arcVersion": "0.0.1-rc.33"
+        "arcVersion": "0.0.1-rc.34"
```
orenyodfat: lets support 34 also for Rinkeby and main net

PR #408, opened 2019-12-09
WIP: Support Competition Scheme

ben-kaufman:


docker-compose.yml
```diff
@@ -18,7 +18,7 @@ services:
       ipfs: 'ipfs:5001'
       ethereum: 'private:http://ganache:8545'
       GRAPH_IPFS_TIMEOUT: '2'
-      GRAPH_MAX_IPFS_FILE_BYTES: '200000'
+      GRAPH_MAX_IPFS_FILE_BYTES: '250000'
```
orenyodfat: increase that to  300000

src/mappings/Competition/mapping.ts
```diff
@@ -0,0 +1,87 @@
+import 'allocator/arena';
```
orenyodfat: redundant

src/mappings/Competition/mapping.ts
```diff
@@ -0,0 +1,87 @@
...
+
+import { BigInt, ByteArray, crypto, ipfs, Address } from '@graphprotocol/graph-ts';
+// Import event types from the Reputation contract ABI
```
orenyodfat: relevant ?

src/mappings/Competition/mapping.ts
```diff
@@ -0,0 +1,87 @@
...
+        event.params._proposalId,
+        event.params._suggestionId as ByteArray,
+      ),
+    ).toHex(),
+  );
+  competitionSuggestion.proposal = event.params._proposalId.toHex();
+  competitionSuggestion.descriptionHash = event.params._descriptionHash;
+  let ipfsData = ipfs.cat('/ipfs/' + event.params._descriptionHash.toHex());
+  if (ipfsData != null) {
+    competitionSuggestion.description = ipfsData.toString();
```
orenyodfat: this should have title as well( same as normal proposal IPFS data).
I guess we should use the same code as in proposal Entity for that. even just call it to parse the ipfs data to `title` and `description`

src/mappings/Competition/mapping.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export function handleRedeem(event: Redeem): void {
+  // TODO: Implement redeem and reward logic.
+}
+
+export function handleNewSuggestion(event: NewSuggestion): void {
+  let competitionSuggestion = new CompetitionSuggestion(
+    crypto.keccak256(
+      concat(
+        event.params._proposalId,
```
orenyodfat: instead event.params._proposalId use contract address

src/mappings/Competition/mapping.ts
```diff
@@ -0,0 +1,87 @@
...
+    competitionSuggestion.description = ipfsData.toString();
+  }
+  competitionSuggestion.suggester = event.params._suggester;
+  competitionSuggestion.totalVotes = BigInt.fromI32(0);
+  competitionSuggestion.createdAt = event.block.timestamp;
+  competitionSuggestion.save();
+}
+
+export function handleNewVote(event: NewVote): void {
+  let suggestionId = crypto.keccak256(
```
orenyodfat: instead event.params._proposalId use contract address

src/mappings/ContributionRewardExt/mapping.ts
```diff
@@ -0,0 +1,187 @@
+import 'allocator/arena';
```
orenyodfat: redundant

src/mappings/ContributionRewardExt/mapping.ts
```diff
@@ -0,0 +1,187 @@
...
+
+import { Address, BigInt, Bytes, crypto, store } from '@graphprotocol/graph-ts';
+
+// Import event types from the Reputation contract ABI
```
orenyodfat: Import event types from the ContributionRewardExt contract ABI

src/mappings/ContributionRewardExt/mapping.ts
```diff
@@ -0,0 +1,187 @@
...
+  ent.descriptionHash = event.params._descriptionHash;
+  ent.externalToken = event.params._externalToken;
+  ent.votingMachine = event.params._intVoteInterface;
+  ent.reputationReward = event.params._reputationChange;
+  let rewards = event.params._rewards;
+  ent.nativeTokenReward = rewards.shift(); // native tokens
+  ent.ethReward = rewards.shift(); // eth
+  ent.externalTokenReward = rewards.shift(); // external tokens
+  ent.periods = BigInt.fromI32(1); // number of periods
+  store.set('ContributionRewardProposal', ent.id, ent);
```
jellegerbrandy: ContributionReward and ContributionRewardExt are different schemes - are you sure you want to use the same entitites here?

For example, the Ext version has fields like `reputationRewardLeft`, where are these going to be stored?

src/mappings/Competition/mapping.ts
ben-kaufman: @jellegerbrandy specification does not include a title property. Are you sure it's part of this here?

src/mappings/ContributionRewardExt/mapping.ts
ben-kaufman: I think so, they're essentially the same, and this is the approach we took with the U/GenericScheme.

src/mappings/Competition/mapping.ts
orenyodfat: yes

src/mappings/ContributionRewardExt/schema.graphql
```diff
@@ -0,0 +1 @@
+
```
orenyodfat: is this empty ?

src/mappings/ContributionRewardExt/mapping.ts
```diff
@@ -0,0 +1,196 @@
...
+  RedeemExternalToken,
+  RedeemNativeToken,
+  RedeemReputation,
+} from '../../types/ContributionRewardExt/ContributionRewardExt';
+
+import * as domain from '../../domain';
+
+import { removeRedeemableRewardOwner } from '../../domain/proposal';
+
+import { shouldRemoveAccountFromUnclaimed, shouldRemoveContributorFromUnclaimed } from '../../domain/reward';
```
orenyodfat: redundant 
as you are using `import * as domain from '../../domain';`

src/mappings/Competition/mapping.ts
jellegerbrandy: I'm sorry I missed that, i changed the issue. A "suggestion"  has a title, a description, and a URL, all optional, and all stored on IPFS, under "descriptionHash"

src/mappings/ContributionRewardExt/schema.graphql
ben-kaufman: Yes like the GenericScheme

src/mappings/ContributionRewardExt/mapping.ts
ben-kaufman: No, this just imports domain/index.ts. Not sufficient.

src/mappings/Competition/mapping.ts
```diff
@@ -64,17 +64,40 @@ export function handleNewSuggestion(event: NewSuggestion): void {
...
     competitionSuggestion.suggester = event.params._suggester;
     competitionSuggestion.totalVotes = BigInt.fromI32(0);
     competitionSuggestion.createdAt = event.block.timestamp;
+    getSuggestionIPFSData(competitionSuggestion);
+
     competitionSuggestion.save();
   }
 }
 
+export function getSuggestionIPFSData(suggestion: CompetitionSuggestion): CompetitionSuggestion {
```
orenyodfat: this is the same as getting proposal ipfs data.
why not use the same code base ?

PR #414, opened 2019-12-11
fix #412

### LINKED ISSUE #412: non-null field error
This is in the `test_env` image, but it is also reproducable by running `npm run migrate` and `npm run deploy` in the current subgraph repository.

The error is this:
```
{
  "errors": [
    {
      "locations": [
        {
          "column": 11,
          "line": 55
        }
      ],
      "message": "Null value resolved for non-null field `genesisProtocolParams`"
    },
    {
      "locations": [
        {
          "line": 87,
          "column": 11
        }
      ],
      "message": "Null value resolved for non-null field `gpQueue`"
    }
  ]
}
```
And the (query is this:
```
    query ProposalState
          {
            proposals {
              ...ProposalFields
              votes {
                id
              }
              stakes {
                id
              }
            }
          }
          fragment ProposalFields on Proposal {
          id
    
          genesisProtocolParams {
            activationTime
            boostedVotePeriodLimit
            daoBountyConst
            limitExponentValue
            minimumDaoBounty
            preBoostedVotePeriodLimit
            proposingRepReward
            queuedVotePeriodLimit
            queuedVoteRequiredPercentage
            quietEndingPeriod
            thresholdConst
            votersReputationLossRatio
          }
          
          gpQueue {
            id
            threshold
            votingMachine
          }
     
        }
```

ben-kaufman:


ops/mappings.json
```diff
@@ -534,20 +534,6 @@
...
       },
-      {
-        "name": "UGenericScheme",
-        "contractName": "UGenericScheme",
-        "dao": "base",
-        "mapping": "UGenericScheme",
-        "arcVersion": "0.0.1-rc.33"
-      },
-      {
-        "name": "SchemeRegistrar",
```
orenyodfat: why this is removed ?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@daostack/subgraph",
-  "version": "0.0.35-0",
+  "version": "0.0.36-0",
```
orenyodfat: why not make it 0.0.35-1 ?

ops/mappings.json
ben-kaufman: It isn't. I just moved it down a bit, git showing it weirdly

package.json
ben-kaufman: Because I try to keep up with the migration repo which is on v36 already

orenyodfat: not sure it is a good must method . and also I do not see that the migration version is incremented on this pr

orenyodfat:
how this pr fix the issue ?
can we think about a test which catch such issues in the future ?

PR #422, opened 2019-12-16
Fix

ben-kaufman:


ben-kaufman:
This is not so basic error, I'm not sure how I could catch it. It happens under very specific conditions not so easy to replicate here.

PR #440, opened 2020-01-09
Competition suggestion extras

### LINKED ISSUE #441: positionInWinnerList is not always correct
Try this query  on https://thegraph.com/explorer/subgraph/jellegerbrandy/daostackjelle . 
This regards a proposal with a rewardsplit [50,50], but some proposals have `positionInWinnerList` higher then `1`  - all of them have `positionInWinnerList` set to a value != null.

I expected all the non-wnning suggestion sto have `positionInWinnerList === null`.

```
query CompetitionSuggestionSearch {
  competitionSuggestions(where: {proposal: "0xb2ee8e9d30db6b0d65bd472f4c0c9e1684f98b8a2b837143ed1c63d944aad86c"}) {
 id
  suggestionId
  proposal {
    id
rewardSplit  }
  
  descriptionHash
  totalVotes
  rewardPercentage
  positionInWinnerList
 }
}
  ```

### PR DESCRIPTION
**Pull request by ben-kaufman**
Fix #441

### DIFF
--- a/src/mappings/Competition/mapping.ts
+++ b/src/mappings/Competition/mapping.ts
@@ -140,32 +140,40 @@ export function handleNewVote(event: NewVote): void {
       suggestion.save();
     }
     vote.save();
-
-    let competition = Competition.bind(event.address);
-    competitionProposal.winningSuggestions = [];
-    let winningSuggestions = competitionProposal.winningSuggestions;
-    debug(competitionProposal.suggestions.toString() + '');
-    for (let i = 0; i < competitionProposal.suggestions.length; i++) {
-      let suggestions = competitionProposal.suggestions;
-      if (suggestions != null) {
-        let currentSuggestionId = (suggestions as string[])[i];
-        let competitionSuggestion = CompetitionSuggestion.load(currentSuggestionId as string);
-        if (competitionSuggestion != null) {
-          let index = competition.getOrderedIndexOfSuggestion(competitionSuggestion.suggestionId);
-          if (index >= competitionProposal.numberOfWinners ||
-            competitionSuggestion.totalVotes.equals(BigInt.fromI32(0))) {
-            competitionSuggestion.positionInWinnerList = null;
-          } else {
-            competitionSuggestion.positionInWinnerList = index;
-            winningSuggestions.push(currentSuggestionId as string);
-          }
-          competitionSuggestion.save();
-        }
-      }
-    }
-    competitionProposal.winningSuggestions = winningSuggestions;
+    let suggestions = competitionProposal.suggestions;
+    sortSuggestions(suggestions as CompetitionSuggestion[]);
     competitionProposal.save();
   }
+}
+
+function sortSuggestions(suggestions: CompetitionSuggestion[]): CompetitionSuggestion[] {
+   let temp: string;
+   for (let i = 0; i < suggestions.length; i++) {
+       let suggestionId = (suggestions as string[])[i];
+       let competitionSuggestion = CompetitionSuggestion.load(suggestionId as string);
+       /*
+        * Place currently selected element array[i]
+        * to its correct place.
+        */
+       for (let j = i + 1; j < suggestions.length ; j++) {
+           let suggestionIdJ = (suggestions as string[])[j];
+           let competitionSuggestionJ = CompetitionSuggestion.load(suggestionIdJ as string);
+           /*
+            * Swap if currently selected array element
+            * is not at its correct position.
+            */
+           if (competitionSuggestion.totalVotes > competitionSuggestionJ.totalVotes) {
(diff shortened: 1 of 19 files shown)

### CODE REVIEW

File: src/mappings/Competition/mapping.ts [hunk 1]
```diff
@@ -100,6 +140,31 @@ export function handleNewVote(event: NewVote): void {
...
     }
     vote.save();
+
+    let competition = Competition.bind(event.address);
+    competitionProposal.winningSuggestions = [];
+    let winningSuggestions = competitionProposal.winningSuggestions;
+    debug(competitionProposal.suggestions.toString() + '');
+    for (let i = 0; i < competitionProposal.suggestions.length; i++) {
+      let suggestions = competitionProposal.suggestions;
+      if (suggestions != null) {
```
jellegerbrandy: I don't get why these lines are _in_ the `for` loop; also do not get how `suggestions` could be `null` without the call to `competitionProposa.suggestions.length` throwing an error...

src/mappings/Competition/mapping.ts
```diff
@@ -100,6 +140,31 @@ export function handleNewVote(event: NewVote): void {
       suggestion.save();
     }
     vote.save();
+
+    let competition = Competition.bind(event.address);
+    competitionProposal.winningSuggestions = [];
+    let winningSuggestions = competitionProposal.winningSuggestions;
+    debug(competitionProposal.suggestions.toString() + '');
```
orenyodfat: does this log is require ? 
please use 'logs' if there is a need to log something.

src/mappings/Competition/mapping.ts
```diff
@@ -100,6 +140,31 @@ export function handleNewVote(event: NewVote): void {
...
+    for (let i = 0; i < competitionProposal.suggestions.length; i++) {
+      let suggestions = competitionProposal.suggestions;
+      if (suggestions != null) {
+        let currentSuggestionId = (suggestions as string[])[i];
+        let competitionSuggestion = CompetitionSuggestion.load(currentSuggestionId as string);
+        if (competitionSuggestion != null) {
+          let index = competition.getOrderedIndexOfSuggestion(competitionSuggestion.suggestionId);
```
orenyodfat: this will return only suggestions within the numberOfWinners.
it will not handle a tie case.

src/mappings/Competition/mapping.ts
```diff
@@ -100,6 +140,31 @@ export function handleNewVote(event: NewVote): void {
...
+          } else {
+            competitionSuggestion.positionInWinnerList = index;
+            winningSuggestions.push(currentSuggestionId as string);
+          }
+          competitionSuggestion.save();
+        }
+      }
+    }
+    competitionProposal.winningSuggestions = winningSuggestions;
+    competitionProposal.save();
```
orenyodfat: if the winningSuggestions is not updated due to the vote so there is no need to re save.

src/mappings/Competition/mapping.ts
```diff
@@ -100,6 +140,31 @@ export function handleNewVote(event: NewVote): void {
       suggestion.save();
     }
     vote.save();
+
+    let competition = Competition.bind(event.address);
+    competitionProposal.winningSuggestions = [];
+    let winningSuggestions = competitionProposal.winningSuggestions;
+    debug(competitionProposal.suggestions.toString() + '');
+    for (let i = 0; i < competitionProposal.suggestions.length; i++) {
```
orenyodfat: how about just sorting the suggestions here ?

src/mappings/Competition/mapping.ts
orenyodfat: it is not optimised to call the contract in a loop for all suggestions.
all this can be avoided by sorting the suggestions.

src/mappings/Competition/mapping.ts
ben-kaufman: No this was left there by mistake

src/mappings/Competition/mapping.ts
ben-kaufman: This should handle ties, why not?

src/mappings/Competition/mapping.ts
ben-kaufman: This will just complexity to save the array and compare.

src/mappings/Competition/mapping.ts
ben-kaufman: I'm not sure sorting is possible at all. If it is then the problem will be supporting tie cases.

src/mappings/Competition/mapping.ts
orenyodfat: no

jellegerbrandy: according to my tests in the client, it handles ties properly

PR #454, opened 2020-01-23
v37-1

ben-kaufman:
Use Arc v39. (Warning: This update breaks compatibility with all versions of the Competition contract older than v39).


test/0.0.1-rc.39/Competition.spec.ts
```diff
@@ -235,7 +236,7 @@ describe('Competition', () => {
 
         await increaseTime(20, web3);
 
-        let suggest = competition.methods.suggest(proposalId, descHash);
+        let suggest = competition.methods.suggest(proposalId, descHash, '0x0000000000000000000000000000000000000000');
```
orenyodfat: how this passed before ?

ben-kaufman: Because it changed only in this version

PR #460, opened 2020-01-29
Fix #459

### LINKED ISSUE #459: add beneficiary field at CompetitionSuggestion entity.
currently suggestion's beneficiary is under `suggester` field.
for now leave  `suggester` field as is to not break alchemy

ben-kaufman:


test/0.0.1-rc.39/Competition.spec.ts
```diff
@@ -289,6 +290,7 @@ describe('Competition', () => {
             url: proposalUrl,
             tags: tagsList,
             suggester: accounts[0].address.toLowerCase(),
+            beneficiary: accounts[0].address.toLowerCase(),
```
jellegerbrandy: the code seems fine,but it would be good to have a test where "suggestor !== beneficiary" just to make sure

ben-kaufman: Added

PR #473, opened 2020-02-11
Patch

ben-kaufman:


src/mappings/Controller/mapping.ts
```diff
@@ -270,106 +280,142 @@ export function handleRemoveGlobalConstraint(
...
+  );
+  let genericSchemeParams = new UGenericSchemeParam(
+    controllerScheme.paramsHash.toHex(),
+  );
+  genericSchemeParams.votingMachine = vmAddress;
+  genericSchemeParams.voteParams = vmParamsHash.toHex();
+  genericSchemeParams.contractToCall = contractToCall;
+  genericSchemeParams.save();
+  controllerScheme.uGenericSchemeParams = genericSchemeParams.id;
+  if (controllerScheme != null) {
```
Jannis: You're accessing `controllerScheme` here before checking whether it is `null`. I've seen the same in a few other places. If `controllerScheme` is null, line 417 may already fail.

PR #474, opened 2020-02-12
Update dorg.json

orishim:
We are migrating all our funds and rep to a new DAO instance.

See migration proposals in our old DAO here:
https://alchemy.daostack.io/dao/0xbe1a98d3452f6da6e0984589e545d4fc25af7526


ben-kaufman:
Hi @orishim , the PR looks fine, just let me know when you are done moving away from the previous DAO so I don't remove it too early.

orishim:
Thanks @ben-kaufman , just had to change the DAO address again. I'll update you when the migration proposal passes.

daos/mainnet/dorg.json
```diff
@@ -1,8 +1,8 @@
 {
-      "name": "dOrg",
-      "Avatar": "0xbe1a98d3452F6da6e0984589e545d4Fc25AF7526",
-      "DAOToken": "0xa2952B15d34d40C9bCd6596A06F9CBB0561236f0",
-      "Reputation": "0xE0A46B46b9F283285b2461b73ebb027eeF39883d",
-      "Controller": "0x809416858a4d0cAA83a660C54B59c4180C6d1Be3",
-      "arcVersion": "0.0.1-rc.16"
-}
+  "arcVersion": "0.0.1-rc.33",
```
orenyodfat: why 33 and not latest 41 ?

orishim:
@ben-kaufman proposals have passed. Funds are now in our new DAO:
https://alchemy.daostack.io/dao/0xd358d4f159e6fae32d1b6096bdace829a5fe33fb

ben-kaufman:
Hi @orishim , please see @orenyodfat comment.

daos/mainnet/dorg.json
dOrgJelli: @orenyodfat @ben-kaufman this is because the DAO was created with the DAOcreator. If DAOcreator moves past v33, adding a Generic Scheme from the UI will no longer be supported. This is because v33 still has a UGenericScheme. If we stop using the UGenericScheme, new DAOs will have to make a PR to the subgraph to add their Generic Schemes.

I think the DAOcreator should wait to upgrade until a solution is created for this issue, either Arc Hives or an intermediate "ad-hoc" solution for tracking newly deployed Generic Schemes.

Issue #476, opened 2020-02-12
add competitionProposal.numWinningSubmissions


dkent600:
Please add `competitionProposal.numWinningSubmissions`.  This will enable us to completely eliminate one query for EthDenver without diminishing UX and associated risky code changes.  Very important.

This `Number` field will provide the count at the current point in time of competitonProposal suggestions having `isWinner` === true.

PR #478, opened 2020-02-13
Competition indexing

### LINKED ISSUE #470: subgraph should supply `CompetitionProposal.admin`
It appears that the subgraph is not supplying `CompetitionProposal.admin` as it should be.  Lacking this value prevents Alchemy from correctly disabling the New Submission button.

This won't compile:

```
{
  proposals {
    id, competition {
      admin
    }
  }
}
```

here:  https://thegraph.com/explorer/subgraph/daostack/v37_3_rinkeby?version=current, which is what Alchemy rinkeby is currently pointed at.

ben-kaufman:
Fix #470 #476


src/mappings/Competition/mapping.ts
```diff
@@ -178,6 +180,7 @@ export function handleNewVote(event: NewVote): void {
     }
     competitionProposal.winningSuggestions = winningSuggestions;
     competitionProposal.totalVotes = competitionProposal.totalVotes.plus(BigInt.fromI32(1));
+    competitionProposal.numWinningSubmissions = BigInt.fromI32(winningSuggestions.length);
     competitionProposal.save();
```
dkent600: we should be consistent here with naming submissions and suggestions

ben-kaufman: But this is the name you asked it to have. So should I change it to `numWinningSuggestions`?

ben-kaufman: I changed it for consistency

src/mappings/Competition/mapping.ts
```diff
@@ -29,7 +29,7 @@ export function handleNewCompetitionProposal(event: NewCompetitionProposal): voi
   competitionProposal.winningSuggestions = [];
   competitionProposal.totalSubmissions = BigInt.fromI32(0);
```
dkent600: should be totalSuggestions

src/mappings/Competition/mapping.ts
dkent600: We are constantly getting confused about "suggestions" and "submissions", hence my error n the request.  Sorry about that.  But subgraph, I think, should itself be consistent.

src/mappings/Competition/schema.graphql
```diff
@@ -16,8 +16,10 @@ type CompetitionProposal @entity {
     winningSuggestions: [CompetitionSuggestion!]
     votes: [CompetitionVote!] @derivedFrom(field: "proposal")
     createdAt: BigInt!
-    totalSubmissions: BigInt!
+    totalSuggestions: BigInt!
     totalVotes: BigInt!
+    numWinningSuggestions: BigInt!
```
dkent600: Perhaps to be consistent with other props here, should be `numberOfWinningSuggestions`

ben-kaufman: ok...

PR #487, opened 2020-02-24
Add Back Old dOrg

dOrgJelli:
We made a mistake :P

We didn't mean to un-index our old DAO, we instead wanted to remove it from the home-page of Alchemy and replace that with our new DAO. To do this we'll make proposals to the DAORegistry instead.

### DIFF
--- a/daos/mainnet/dorg.json
+++ b/daos/mainnet/dorg.json
@@ -1,8 +1,8 @@
 {
-  "arcVersion": "0.0.1-rc.33",
   "name": "dOrg",
-  "Avatar": "0xd358D4F159E6FAE32d1B6096bDaCe829a5fe33fb",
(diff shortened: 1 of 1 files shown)


dOrgJelli:
Requesting registry update here: https://github.com/daostack/arc-hive/issues/21

ben-kaufman:
Hi @dOrgJelli , so if you want to have both DAOs indexed you'll need to have them in 2 files instead of overwriting the existing one. Please add the older DAO in a separate file so that you'll have both.

dOrgJelli:
@ben-kaufman the new DAO is already indexed because it was created with the DAOcreator.sol, and the subgraph was updated to index all newly created DAOs from there. You can see the new DAO in Alchemy here: https://alchemy.daostack.io/dao/0xd358d4f159e6fae32d1b6096bdace829a5fe33fb

PR #494, opened 2020-03-12
Signal Scheme

mrrobot16:


ops/mappings.json
```diff
@@ -8,6 +8,13 @@
         "mapping": "DAOToken",
         "arcVersion": "0.0.1-rc.16"
       },
+      {
+        "name": "NewSignalScheme",
+        "dao": "address",
+        "mapping": "SignalScheme",
```
orenyodfat: I thought we are using GenericScheme ?
is that needed ?

orenyodfat: @cryptogoth ?

cryptogoth: Hi @orenyodfat I've never seen that scheme before, and haven't delved into daostack repo much :)

It sounds cool though.

PR #498, opened 2020-03-16
Subgraph v5 experimental

ben-kaufman:


src/mappings/Controller/schema.graphql
```diff
@@ -87,6 +89,13 @@ type ContributionRewardExtParam @entity {
 	rewarder: Bytes!
 }
 
+type SchemeFactoryParam @entity {
+	id: ID!
+	votingMachine: Bytes!
+	voteParams: GenesisProtocolParam!
+	daoFactory: Bytes!
```
orenyodfat: why daoFactory is needed here ?

src/mappings/SchemeFactory/mapping.ts
```diff
@@ -0,0 +1,101 @@
...
+// Import entity types generated from the GraphQL schema
+import {
+   SchemeFactoryNewSchemeProposal,
+   SchemeFactoryProposal,
+   SchemeFactoryProposalExecuted,
+} from '../../types/schema';
+import { equalsBytes, equalStrings, eventId } from '../../utils';
+
+export function handleNewSchemeProposal(event: NewSchemeProposal): void {
+  let ent = SchemeFactoryNewSchemeProposal.load(eventId(event));
```
orenyodfat: how about the id will be the proposalId?

src/mappings/Controller/schema.graphql
ben-kaufman: It's a parameter of the SchemeFactory. It's not necessary I guess but I can't l can't know what will be needed by the UI team and we usually index all parameters like that.

src/mappings/SchemeFactory/mapping.ts
```diff
@@ -0,0 +1,101 @@
...
+                          ent.schemeName,
+                          ent.schemeData,
+                          ent.packageVersion as BigInt[],
+                          ent.permission,
+                          ent.schemeToReplace as Address);
+}
+
+export function handleProposalExecuted(event: ProposalExecuted): void {
+  let ent = SchemeFactoryProposalExecuted.load(eventId(event));
+  if (ent == null) {
```
orenyodfat: is it not always null ?

src/mappings/SchemeFactory/mapping.ts
ben-kaufman: That's how we did it in the SchemeRegistrar and other schemes. We can change it I guess but didn't want to make it" stand out".

src/mappings/SchemeFactory/mapping.ts
ben-kaufman: I'm pretty sure you have to have these checks even when they seem futile, if I remember correctly that's required due to some Graph Node special requirements.

src/mappings/SchemeFactory/mapping.ts
orenyodfat: well:
1. event is unique and for that this line is redundant if we use event
2. I feel that proposalId is better

ben-kaufman: ok...

src/mappings/SchemeFactory/mapping.ts
orenyodfat: if that is the case so it is an issue with the graph , which give same event id for multiple events.
using porposalId is better..

PR #503, opened 2020-03-26
Migration v8 experimental

ben-kaufman:


src/mappings/JoinAndQuit/schema.graphql
```diff
@@ -0,0 +1,9 @@
+type JoinAndQuitProposal @entity {
+	id: ID!
+	dao: DAO!
+	proposedMember: Bytes!
+	funder: Bytes!
+	funding: BigInt!
+	executed: Boolean!
```
orenyodfat: might need addition filed to say if the dao accept or reject the proposal.

src/mappings/JoinAndQuit/mapping.ts
```diff
@@ -0,0 +1,56 @@
+import {
+  JoinInProposal,
+  ProposalExecuted,
+  RedeemReputation,
+} from '../../types/JoinAndQuit/JoinAndQuit';
+
```
orenyodfat: where is rageQuit is handled ?

src/mappings/JoinAndQuit/datasource.yaml
```diff
@@ -0,0 +1,18 @@
...
+entities:
+  - JoinAndQuitProposal
+eventHandlers:
+  - event: ProposalExecuted(indexed address,indexed bytes32,int256)
+    handler: handleProposalExecuted
+  - event: JoinInProposal(indexed address,indexed bytes32,string,address,uint256)
+    handler: handleNewJoinAndQuitProposal
+  - event: FundedBeforeDeadline(indexed address)
+    handler: handleFundedBeforeDeadline
+  - event: Donation(indexed address,indexed uint256)
```
orenyodfat: There is no Donation event .

src/mappings/JoinAndQuit/schema.graphql
```diff
@@ -0,0 +1,9 @@
...
+	reputationMinted: BigInt!
```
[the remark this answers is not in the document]
ben-kaufman: yes

src/mappings/JoinAndQuit/schema.graphql
ben-kaufman: We have that in the Proposal entity. We could also remove the executed I guess.

src/mappings/JoinAndQuit/mapping.ts
ben-kaufman: What should the subgraph do in a ragequit?

src/mappings/JoinAndQuit/datasource.yaml
ben-kaufman: Removed

src/mappings/JoinAndQuit/mapping.ts
orenyodfat: index the event of that and for each funder decreased it funding.
does a ragequit member should get out of the dao is a good point.

src/mappings/FundingRequest/schema.graphql
```diff
@@ -0,0 +1,8 @@
+type FundingRequestProposal @entity {
+	id: ID!
```
orenyodfat: should it include also if its token or native coin ?
if token , so its address ?

ben-kaufman: It already includes it in the scheme parameters. No need to cache this per proposal as it is the same for all proposals of the scheme

src/mappings/JoinAndQuit/mapping.ts
ben-kaufman: I will add that when you release the Arc v9 with the fix of the ragequit

orenyodfat: do you mean arc V11 ?

PR #504, opened 2020-03-29
Fixes For test_env-experimental

dOrgJelli:


ben-kaufman:
Hi, looks good, just please remove the change to the version. The subgraph version is not coupled with the migration version and changes according to the changes to the subgraph behavior itself.

dOrgJelli:
@jellegerbrandy says that the arc->migration->subgraph versions should be aligned

PR #513, opened 2020-04-06
Use migration v12

ben-kaufman:


test/0.1.1-rc.12/domain.spec.ts
```diff
@@ -28,7 +28,7 @@ describe('Domain Layer', () => {
   let accounts;
   const orgName = getOrgName();
   const tokenName = orgName + ' Token';
-  const tokenSymbol = orgName[0] + orgName.split(' ')[1][0] + 'T';
+  const tokenSymbol = 'MY';
```
jellegerbrandy, line 31: why this change?

ben-kaufman: Because Jordan changed the DAO name and the token name and symbol to be constant

PR #530, opened 2020-04-27
Create fit-together.json

ben-kaufman:


daos/mainnet/fit-together.json
```diff
@@ -0,0 +1,34 @@
...
+        "name": "ReputationAdmin",
+        "alias": "ReputationAdmin",
+        "address": "0x550F422C14780f64685A0959da12b528531847BA",
+        "arcVersion": "0.0.1-rc.41"
+      }
+    ],
+    "StandAloneContracts": [
+      {
+        "name": "ContributionRewardExt",
+        "address": "0xD598aAd8700bA3c588c3A2d3A43de6C5D69De23b",
```
orenyodfat, line 24: why there are 2 ContributionRewardExt?

daos/mainnet/fit-together.json
```diff
@@ -0,0 +1,34 @@
...
+      {
+        "name": "ContributionRewardExt",
+        "alias": "ContributionRewardExt",
+        "address": "0xD598aAd8700bA3c588c3A2d3A43de6C5D69De23b",
+        "arcVersion": "0.0.1-rc.41"
+      },
+      {
+        "name": "ReputationAdmin",
+        "alias": "ReputationAdmin",
+        "address": "0x550F422C14780f64685A0959da12b528531847BA",
```
orenyodfat, line 17: the owner of this is 0x06044b5359d8df7886366c22c61c7ecd29becac7 .
who is it ?

daos/mainnet/fit-together.json
ben-kaufman: There is only one (note the address is the same for both).
That's "duplicated" because of how the migration deals with the Contribution Reward + Competition interdependency.
It's like that in all daos, see for example https://github.com/daostack/subgraph/blob/master/daos/mainnet/cryptochicks.json

daos/mainnet/fit-together.json
ben-kaufman: I don't know, @ Theylon should know that.

PR #547, opened 2020-05-08
dao metadata support

### LINKED ISSUE #545: index dao metadata
the event come from : 
https://github.com/daostack/arc/blob/arc-factory/contracts/controller/Avatar.sol#L169

orenyodfat:
resolve #545

### DIFF
--- a/test/0.1.1-rc.17/util.ts
+++ b/test/0.1.1-rc.17/util.ts
@@ -100,6 +100,13 @@ export async function writeProposalIPFS(data: any) {
   return ipfsResponse[0].path;
 }
 
+export async function writeToIPFS(data: string) {
+  const ipfsClient = IPFSClient(ipfs);
+  const ipfsResponse = await ipfsClient.add(new Buffer(data));
+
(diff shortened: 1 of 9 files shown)


test/0.1.1-rc.17/util.ts
```diff
@@ -100,6 +100,13 @@ export async function writeProposalIPFS(data: any) {
   return ipfsResponse[0].path;
 }
 
+export async function writeToIPFS(data: string) {
```
ben-kaufman, line 103: We already have this though (but with stringify, does it mater?): https://github.com/daostack/subgraph/blob/a3a63320d93e513782700fd62c3d65288f62db31/test/0.1.1-rc.17/util.ts#L96

orenyodfat: `writeProposalIPFS`  this is doing JSON.stringify ...which is different.

ben-kaufman: ok

PR #553, opened 2020-05-12
Add lastUpdatedAt

ben-kaufman:


src/utils.ts
```diff
@@ -124,3 +125,11 @@ export function fixJsonQuotes(target: string): string {
      }
      return result;
 }
+
+export function save(ent: Entity | null, entName: string, timestamp: BigInt): void {
+  if (ent == null) {
+    return;
+  }
+  (ent as Entity).setBigInt('lastUpdatedAt', timestamp);
+  store.set(entName, ent.getString('id'), ent as Entity);
```
orenyodfat, line 134: does ent.save () will not do the work here ?

src/utils.ts
```diff
@@ -124,3 +125,11 @@ export function fixJsonQuotes(target: string): string {
      }
      return result;
 }
+
+export function save(ent: Entity | null, entName: string, timestamp: BigInt): void {
```
orenyodfat, line 129: why it need to get entName ?

src/utils.ts
ben-kaufman: No, it doesn't exist here

src/utils.ts
ben-kaufman: Becuase  ent.save() doesn't exist

ben-kaufman: I know this need for the `entName ` is bad and very very error-prone, but I don't think there is a better way around it.

orenyodfat: we might missing the asserts that .save do ..
e.g 
`assert(id !== null, "Cannot save AvatarContract entity without an ID");
    assert(
      id.kind == ValueKind.STRING,
      "Cannot save AvatarContract entity with non-string ID. " +
        'Considering using .toHex() to convert the "id" to a string.'
    );`

ben-kaufman: I guess we could manually add them, but this is obviously not a very optimal solution...

PR #559, opened 2020-05-18
Sokol + xDAI

ben-kaufman:


ops/generate-contractsinfo.js
```diff
@@ -1,5 +1,8 @@
 const fs = require("fs");
-const { migrationFileLocation: defaultMigrationFileLocation, network } = require("./settings");
+let { migrationFileLocation: defaultMigrationFileLocation, network } = require("./settings");
+if (network === "poa-sokol") {
+  network = "sokol"
```
orenyodfat, line 4: why not just call that poa-sokol ..?

ben-kaufman: We'll need to change that in the migration then too which means another release for it when anyway I'd prefer `sokol` to `poa-sokol`

PR #575, opened 2020-06-21
V8 6

ben-kaufman:


src/mappings/DAOFactory/mapping.ts
```diff
@@ -84,6 +85,14 @@ export function handleProxyCreated(event: ProxyCreated): void {
   // Ensure the FactoryContract has been added to the store
   getDAOFactoryContract(event.address);
 
+  // Don't create contract if the proxy admin is the default address
+  let proxy = AdminUpgradeabilityProxy.bind(event.params._proxy);
+  let callResult = proxy.try_admin();
```
orenyodfat, line 90: here it assume that the caller is admin.  as only admin can call admin(). https://github.com/OpenZeppelin/openzeppelin-sdk/blob/master/packages/lib/contracts/upgradeability/BaseAdminUpgradeabilityProxy.sol#L45 .
Not sure it is a good assumption.

ben-kaufman: It exactly assumes that, because if it is the admin we don’t want to index it, that’s how it should to work

PR #578, opened 2020-07-13
Add UniversalContractInfo entity

ben-kaufman:
Add UniversalContractInfo entity to keep all versions of Genesis Protocol and Redeemer.


orenyodfat:
please add test

ben-kaufman:
Added

PR #582, opened 2020-07-23
Use Arc v43

ben-kaufman:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@daostack/subgraph",
-  "version": "0.0.39-5",
+  "version": "0.0.39-6",
```
orenyodfat, line 3: did we deployed 39-5?

ben-kaufman: We released it... https://github.com/daostack/subgraph/releases/tag/ganache-0.0.1-rc.41-v7-0.0.39-5

PR #584, opened 2020-07-26
Improve balance tracking

ben-kaufman:


src/mappings/GenericScheme/mapping.ts
```diff
@@ -47,6 +48,9 @@ export function handleProposalExecuted(
   if (ent != null) {
     ent.executed = true;
     ent.returnValue = event.params._genericCallReturnValue;
+    if (BigInt.fromI32(0).lt(ent.value)) {
```
orenyodfat, line 51: should it check if the call succeed?

ben-kaufman: Well if the call fails the event won't happen in the first place. So I don't think we need to/ can do that.

orenyodfat: not sure about that as if https://github.com/daostack/arc/blob/master/contracts/controller/Avatar.sol#L58 revert the event will still be emitted

ben-kaufman: No, we check here before calling the Executed event that the call succeeded https://github.com/daostack/arc/blob/1d95202c3d2c2daef7c8dffcf7935e236c15c24a/contracts/schemes/GenericScheme.sol#L114

orenyodfat: I think this should be handle at the Avatar module listen to the handle GenericCall event.
not in the genericScheme

ben-kaufman: Ok, I changed it

src/mappings/Avatar/mapping.ts
```diff
@@ -27,6 +28,12 @@ function handleAvatarBalance(
   store.set('AvatarContract', avatar.id, avatar);
 }
 
+export function handleGenericCall(event: GenericCall): void {
+  if (event.params._success) {
```
orenyodfat, line 32: should it also check if the balance is not 0 ?

PR #586, opened 2020-08-03
Token trade scheme

cbrzn:


ops/mappings.json
```diff
@@ -133,6 +133,13 @@
           "dao": "package",
           "mapping": "DAOFactory",
           "arcVersion": "0.1.2-rc.1"
+        },
+        {          
+          "name": "TokenTrade",
+          "contractName": "TokenTrade",
+          "dao": "package",
+          "mapping": "TokenTrade",
+          "arcVersion": "0.1.2-rc.2"
```
cbrzn, line 142: This version might change

test/0.1.2-rc.0/util.ts
```diff
@@ -69,6 +69,7 @@ export function getContractAddresses() {
     SchemeFactory: addresses.private.dao[arcVersion].Schemes[5].address,
     JoinAndQuit: addresses.private.dao[arcVersion].Schemes[6].address,
     FundingRequest: addresses.private.dao[arcVersion].Schemes[7].address,
+    // TokenTrade: addresses.private.dao[arcVersion].Schemes[8].address,
```
ben-kaufman, line 72: You'll need to create the TokenTrade scheme in the test itself with `SchemeFactory` as it is not included in the test dao like other schemes (can't add so many schemes at the same time...). You should be able to do this fairly easily by following the example of the `SchemeFactory` test to see how to propose and pass in voting a proposal for a new scheme.

cbrzn: All right, if I'm not wrong we are still getting the TokenTrade address from this function, but it will live inside the object `...addresses.private.package[arcVersion]` (like 60)

ops/mappings.json
ben-kaufman: Yes. Also please make sure to add this to Ganache, Rinkeby, Kovan, xDai.

ben-kaufman:
I have changed the branch to use the new migration, please rebase.

ops/generate-subgraph.js
```diff
@@ -66,9 +66,13 @@ function combineFragments(fragments, isTemplate, addresses, missingAddresses) {
       eventHandlers = yamlLoad.eventHandlers;
       entities = yamlLoad.entities;
       abis = (yamlLoad.abis || [contract]).map(contractName => {
+        let correctedVersion = version;
+        if(contractName == 'TokenTrade' && parseInt(version.slice(-1)[0]) <= 3) {
```
cbrzn, line 70: This is because when this script runs, it generates in the `subgraph.yaml` the following entry in the `abis` of controller:
```
path/ops/../abis/0.1.2-rc.0/TokenTrade.json
```
Causing an error, because the TokenTrade is after the version `.3` - If there's a better solution please, let us know @ben-kaufman

cbrzn:
Test is working now! Please fill free to review the work sir @ben-kaufman

ben-kaufman:
Hi, great, please just switch the order of schemes in the DAOFactory test so it passes too.

package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "@daostack/subgraph-experimental",
-  "version": "0.1.1-rc.8-v10",
+  "name": "@dorgtech/subgraph-experimental",
```
ben-kaufman, line 2: Please remove this change...

ben-kaufman: Also, why not v11?

release.sh
```diff
@@ -35,15 +35,15 @@ fi
 
 # commit the postgres image
 container_id=$(docker ps  -f "name=postgres" -l -q)
-image_name=daostack/subgraph-experimental-postgres
+image_name=dorgtech/subgraph-experimental-postgres
```
ben-kaufman, line 38: Change back as well

package.json
cbrzn: Because we were deploying it multiple time in order to release the `test_env` and test arc, which is also done and tested :-)

release.sh
```diff
@@ -58,8 +58,8 @@ npm install
 npm login
 npm publish
 # tag on github
-echo "create tag ${image_version}"
-git tag -a $image_version -m "Release of version $image_name:$image_version"
-git push --tags
+# echo "create tag ${image_version}"
```
ben-kaufman, line 61: Please just return this

PR #589, opened 2020-08-10
Upgrade to Arc v6 experimental

ben-kaufman:


ops/generate-subgraph.js
```diff
@@ -67,7 +67,10 @@ function combineFragments(fragments, isTemplate, addresses, missingAddresses) {
       entities = yamlLoad.entities;
       abis = (yamlLoad.abis || [contract]).map(contractName => {
         let correctedVersion = version;
-        if(contractName == 'TokenTrade' && parseInt(version.slice(-1)[0]) <= 3) {
+        if(contractName === 'Join' && parseInt(version.slice(-1)[0]) <= 5) {
+          correctedVersion = correctedVersion.substring(0, correctedVersion.length -1) + 6
+        }
+        if(contractName === 'TokenTrade' && parseInt(version.slice(-1)[0]) <= 3) {
```
orenyodfat, line 73: why this is needed ?

ops/generate-subgraph.js
```diff
@@ -155,18 +158,20 @@ function buildTemplates() {
...
-          name,
-          mapping,
-          arcVersion
-        }],
-        true,
-        undefined,
-        undefined
-      )
-    );
+    if (arcVersion !== '0.1.2-rc.5') {
```
orenyodfat, line 161: why this if is needed?

ops/generate-subgraph.js
ben-kaufman: Because these schemes don't exist in older versions, but the controller is using them to index their parameters

ops/generate-subgraph.js
ben-kaufman: Because we don't have an Arc rc.5... (well not a valid release at least)

ops/generate-subgraph.js
orenyodfat: can we avoid it and make the controller not using them?

ops/generate-subgraph.js
orenyodfat: so why add that to the code here?

ops/generate-subgraph.js
ben-kaufman: No, not if we want to index their parameters at lease

ops/generate-subgraph.js
ben-kaufman: To skip on it, because the loop will try to create for it as well

orenyodfat: you need to skip it on migration.

PR #591, opened 2020-08-19
Add CurveLabs necDAO for test purposes to rinkeby DAOs

osarrouy:
Hey,

Here is the PR to demo our Uniswap integration to our clients at necDAO.

The only thing I'm not sure of is: the `UniswapProxy` scheme comes with no specific frontend. Users will only interact with it through the `GenericScheme`. It's still registered as a scheme because it has the permission to call `genericCall` on the `Avatar` to `approve` ERC20 transfers.

I've added it to this file anyhow so that it displays a human-readable name instead of a contract address in the Alchemy schemes list. This will not cause a bug, right ?

Thanks for reviewing the PR anyhow.

### DIFF
--- a/daos/rinkeby/necdaocurvelabs.json
+++ b/daos/rinkeby/necdaocurvelabs.json
@@ -1,0 +1,23 @@
+{
+  "name": "necDAO",
+  "Avatar": "0x331dDC7Ed27451b5B4d25A41e8dF53cFF456DcA9",
+  "DAOToken": "0x6939C5672b575C2fa4E3393D1706c7A097219805",
+  "Reputation": "0x814d49253d595Ec7341e7307d2CB4964Ab9655F6",
+  "Controller": "0x5bdD6bdb6D35920DC1DeD2101Cb9b1D39DCEa92D",
+  "Schemes": [
+    {
+      "name": "UniswapProxy",
+      "alias": "UniswapProxy",
+      "address": "0x50AE25C926176D71F41A27Fe97fa4A7c31207131",
+      "arcVersion": "0.0.1-rc.44"
+    },
+    {
+      "name": "GenericScheme",
+      "alias": "Uniswap",
+      "address": "0xc3524072aa711f84d1Db93B1cdad68be176162bd",
+      "arcVersion": "0.0.1-rc.44"
+    }
+  ],
+  "StandAloneContracts": [],
+  "arcVersion": "0.0.1-rc.44"
+}


daos/rinkeby/necdaocurvelabs.json
```diff
@@ -0,0 +1,23 @@
...
+  "name": "necDAO",
+  "Avatar": "0x331dDC7Ed27451b5B4d25A41e8dF53cFF456DcA9",
+  "DAOToken": "0x6939C5672b575C2fa4E3393D1706c7A097219805",
+  "Reputation": "0x814d49253d595Ec7341e7307d2CB4964Ab9655F6",
+  "Controller": "0x5bdD6bdb6D35920DC1DeD2101Cb9b1D39DCEa92D",
+  "Schemes": [
+    {
+      "name": "UniswapProxy",
+      "alias": "UniswapProxy",
+      "address": "0x50AE25C926176D71F41A27Fe97fa4A7c31207131",
```
orenyodfat, line 11: is this scheme can be verify that on [etherscan](https://rinkeby.etherscan.io/address/0x50AE25C926176D71F41A27Fe97fa4A7c31207131#code) ? it is missing the implementation of of it.  [see](https://rinkeby.etherscan.io/proxyContractChecker?a=0x50AE25C926176D71F41A27Fe97fa4A7c31207131)

daos/rinkeby/necdaocurvelabs.json
```diff
@@ -0,0 +1,23 @@
...
+      "arcVersion": "0.0.1-rc.44"
+    },
+    {
+      "name": "GenericScheme",
+      "alias": "Uniswap",
+      "address": "0xc3524072aa711f84d1Db93B1cdad68be176162bd",
```
orenyodfat, line 17: it will be nice to verify [this](https://rinkeby.etherscan.io/address/0xc3524072aa711f84d1Db93B1cdad68be176162bd#code) also

orenyodfat:
`I've added it to this file anyhow so that it displays a human-readable name instead of a contract address in the Alchemy schemes list. This will not cause a bug, right ?`

I guess it is ok. where we can find the UniswapProxy code ?

daos/rinkeby/necdaocurvelabs.json
osarrouy: This is a `GenericScheme` contract. I can't verify it because it has been automatically deployed through the `@daostack/migration` package and thus I don't have access to the exact `constructor` params - notably the voting params hash.

osarrouy:
The repository for the source code is [here](0xF4D96486b20CF4D27e5f95a4dB6CDc274Dc1404E).

daos/rinkeby/necdaocurvelabs.json
osarrouy: This contract is now verified and registered as Proxy so people can access its implementation. I am the current `admin` of this proxy. I'm not sure we're gonna stick to this proxy architecture for the mainnet release but the point here was to be able to incrementally update the proxy if requested by the client with no need to update the subgraph again [cause the indexed address would stay the same]. In case we use the same architecture on mainnet, of course, the admin rights would be granted to the DAO :)

orenyodfat: preferable to not have that as a proxy for production

orenyodfat:
> The repository for the source code is [here](0xF4D96486b20CF4D27e5f95a4dB6CDc274Dc1404E).

wrong link ? @osarrouy

osarrouy:
Damned. I updated the original link :)

daos/rinkeby/necdaocurvelabs.json
orenyodfat: ok. I verified it :)

orenyodfat (approved):

ben-kaufman (approved):

daos/rinkeby/necdaocurvelabs.json
dkent600: > preferable to not have that as a proxy for production

I am curious to understand why not?

PR #597, opened 2020-09-14
Draft: Added GenericSchemeMultiCall

nicoelzer:
- Added GenericSchemeMultiCall


daos/rinkeby/dxdao.json
```diff
@@ -13,4 +13,4 @@
     }
   ],
   "arcVersion": "0.0.1-rc.19"
-}
+}
```
orenyodfat, line 16: please remove this update

src/domain/proposal.ts
```diff
@@ -392,6 +392,24 @@ export function updateGSProposal(
   saveProposal(proposal);
 }
 
+export function updateGSMCProposal(
```
orenyodfat, line 395: is that not the same as `updateGSProposal`

orenyodfat (changes requested):
please add tests

PR #605, opened 2020-10-21
Add vault to dao object

ben-kaufman:


src/domain/dao.ts
```diff
@@ -6,6 +6,13 @@ export function getDAO(id: string): DAO {
   let dao = store.get('DAO', id) as DAO;
   if (dao == null) {
     dao = new DAO(id);
+    dao.name = '';
```
orenyodfat, line 9: why is this needed?

ben-kaufman: because the vault event happens before the dao creation event, and if it will be null the sbgraph will crash as it is a non null field

orenyodfat (approved):

src/mappings/DAOFactory/mapping.ts
```diff
@@ -99,7 +99,11 @@ export function handleProxyCreated(event: ProxyCreated): void {
 
   if (equalStrings(event.params._contractName, 'Avatar')) {
     let avatar = Avatar.bind(event.params._proxy);
-    addContract(avatar.vault(), 'Vault', version);
+    let vault = avatar.vault();
+    addContract(vault, 'Vault', version);
+    let dao = getDAO(event.params._proxy.toHex());
```
orenyodfat, line 104: `getDAO` will create new empty dao if it not exist.  which is not what we want.
the code should add vault only to already exist `Dao` entity

PR #610, opened 2020-10-27
Add unregistered proposals counting

### LINKED ISSUE #608: add NumberOfBoostedProposals of unregistered scheme

ben-kaufman:
Close #608

### DIFF
--- a/src/mappings/Controller/mapping.ts
+++ b/src/mappings/Controller/mapping.ts
@@ -88,6 +88,34 @@ function unregisterScheme(avatarAddress: Address, scheme: Address): void {
   let controllerScheme = ControllerScheme.load(crypto.keccak256(concat(avatarAddress, scheme)).toHex());
   if (controllerScheme != null) {
     controllerScheme.isRegistered = false;
+    let dao = DAO.load(avatarAddress.toHex());
+    if (dao != null) {
+      dao.numberOfQueuedProposals = dao.numberOfQueuedProposals.minus(
+        controllerScheme.numberOfQueuedProposals,
+      );
+      dao.numberOfPreBoostedProposals = dao.numberOfPreBoostedProposals.minus(
+        controllerScheme.numberOfPreBoostedProposals,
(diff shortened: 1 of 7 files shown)


src/mappings/Controller/mapping.ts
```diff
@@ -88,6 +88,34 @@ function unregisterScheme(avatarAddress: Address, scheme: Address): void {
   let controllerScheme = ControllerScheme.load(crypto.keccak256(concat(avatarAddress, scheme)).toHex());
   if (controllerScheme != null) {
     controllerScheme.isRegistered = false;
+    let dao = DAO.load(avatarAddress.toHex());
+    if (dao != null) {
+      dao.numberOfQueuedProposals = dao.numberOfQueuedProposals.minus(
```
orenyodfat, line 93: this cannot be negative because of https://github.com/daostack/arc/pull/802 . right ?

src/mappings/Controller/mapping.ts
```diff
@@ -136,6 +164,36 @@ export function handleRegisterScheme(event: RegisterScheme): void {
     return;
   }
 
+  let dao = DAO.load(avatar.toHex());
+  let controllerScheme = ControllerScheme.load(crypto.keccak256(concat(avatar, event.params._scheme)).toHex());
+  if (dao != null && controllerScheme != null) {
+    dao.numberOfQueuedProposalsUnregistered = dao.numberOfQueuedProposalsUnregistered.minus(
```
orenyodfat, line 170: is there a scenario that this can be negative?

ben-kaufman: Yes if the scheme is re-registered

src/mappings/Controller/mapping.ts
ben-kaufman: No this happens when a scheme is unregistered and still have proposals on it

orenyodfat (approved):

PR #620, opened 2020-12-06
Add scheme error field

ben-kaufman:
Add error is universal scheme params were non-existent.

### DIFF
--- a/src/domain/gpqueue.ts
+++ b/src/domain/gpqueue.ts
@@ -128,6 +128,6 @@ export function create(dao: Address,
                       gpAddress,
                       BigInt.fromUnsignedBytes(bigOne as Bytes),
                       organizationId as Bytes,
-                      crypto.keccak256(concat(dao, scheme)).toHex());
+                      schemeId);
    }
 }
--- a/src/domain/gpqueue.ts
+++ b/src/domain/gpqueue.ts
@@ -45,6 +45,7 @@ export function create(dao: Address,
    if (contractInfo ==  null) {
      return;
    }
+   let schemeId = crypto.keccak256(concat(dao, scheme)).toHex();
    let gpAddress: Address;
    let isGPQue = false;
    let addressZero = '0x0000000000000000000000000000000000000000';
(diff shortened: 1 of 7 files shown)


src/utils.ts
```diff
@@ -133,3 +135,14 @@ export function fixJsonQuotes(target: string): string {
      }
      return result;
 }
+
+export function setSchemeParamsError(dao: Address, scheme: Address): void {
```
orenyodfat, line 139: it might be cleaner to call it with the schemeId .

src/domain/gpqueue.ts
```diff
@@ -55,6 +55,8 @@ export function create(dao: Address,
        gpAddress = parameters.value1;
        setContributionRewardParams(dao, scheme, gpAddress, parameters.value0);
        isGPQue = true;
+     } else {
+       setSchemeParamsError(dao, scheme);
```
orenyodfat, line 59: it might be cleaner to get the schemeId earlier .
use it also for 
https://github.com/daostack/subgraph/blob/11e36b95e14f66867f6b1290d39f88b006ec79c5/src/domain/gpqueue.ts#L125
and call setSchemeParamsError with schemeId

src/domain/gpqueue.ts
```diff
@@ -56,7 +56,7 @@ export function create(dao: Address,
        setContributionRewardParams(dao, scheme, gpAddress, parameters.value0);
        isGPQue = true;
      } else {
-       setSchemeParamsError(dao, scheme);
+       setSchemeParamsError(crypto.keccak256(concat(dao, scheme)).toHex());
```
orenyodfat, line 59: the schemeId can be get on the beginning of the function and rep in local var : 
let schemeId = crypto.keccak256(concat(dao, scheme)).toHex()

src/domain/gpqueue.ts
```diff
@@ -45,6 +45,7 @@ export function create(dao: Address,
    if (contractInfo ==  null) {
      return;
    }
+   let schemeId = crypto.keccak256(concat(dao, scheme)).toHex();
```
orenyodfat, line 48: schemeId can be use also in line 126

src/utils.ts
```diff
@@ -133,3 +135,12 @@ export function fixJsonQuotes(target: string): string {
      }
      return result;
 }
+
+export function setSchemeParamsError(schemeId: string): void {
+  let controllerScheme = ControllerScheme.load(schemeId);
+  if (controllerScheme != null) {
+    controllerScheme.error = 'Scheme parameters could not be found.';
+    controllerScheme.save();
+  }
```
orenyodfat, line 144: else

orenyodfat (approved):

orenyodfat (approved):

PR #622, opened 2020-12-13
v41-3

ben-kaufman:
- Use Arc v52
- Increase closingAt time by 1500000000 after execution to make it very high.

### DIFF
--- a/src/domain/proposal.ts
+++ b/src/domain/proposal.ts
@@ -437,7 +437,10 @@ export function updateProposalExecution(
 ): void {
   let proposal = getProposal(proposalId.toHex());
   proposal.executedAt = timestamp;
-  proposal.closingAt = timestamp.plus(BigInt.fromI32(1500000000));
+  // Setting the closingAt field to a far away point in the future so it will be easy to 
+  // sort all proposal(open and executed) in ascending order by the closingAt field
+  const CLOSING_AT_TIME_INCREASE = 1500000000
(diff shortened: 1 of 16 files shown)


src/domain/proposal.ts
```diff
@@ -437,7 +437,7 @@ export function updateProposalExecution(
 ): void {
   let proposal = getProposal(proposalId.toHex());
   proposal.executedAt = timestamp;
-  proposal.closingAt = timestamp;
+  proposal.closingAt = timestamp.plus(BigInt.fromI32(1500000000));
```
orenyodfat, line 440: can we make is a const somehow ? 
add a comment : `Setting the closingAt field to a far away point in the future so it will be easy to sort all proposal(open and executed) in ascending order by the closingAt field`

src/domain/proposal.ts
```diff
@@ -437,7 +437,10 @@ export function updateProposalExecution(
 ): void {
   let proposal = getProposal(proposalId.toHex());
   proposal.executedAt = timestamp;
-  proposal.closingAt = timestamp;
+  // Setting the closingAt field to a far away point in the future so it will be easy to 
+  // sort all proposal(open and executed) in ascending order by the closingAt field
+  const CLOSING_AT_TIME_INCREASE = 1500000000
```
orenyodfat, line 442: you can move it to some config file and use the same const in the test.. no ?

orenyodfat (approved):

